PR #32, opened 2021-07-20
display transaction logs on transaction details

AlexandruPislariu:


tanghel (commented):
In the SC Results object, also include logs attribute

tanghel (approved):

PR #46, opened 2021-07-30
add config legacy elastic

CristianFodor:
Add config value for Legacy Elastic


tanghel:
This will be fixed when we merge the dev branch in the main branch. we only reference one elastic at the moment

PR #82, opened 2021-08-17
Apr detailed

tanghel:


constantintovisi (approved):

src/endpoints/network/network.service.ts
```diff
@@ -201,4 +166,61 @@ export class NetworkService {
...
+    let config = await this.getNetworkConfig();
+    let stake = await this.stakeService.getGlobalStake();
+    const { account: { balance: stakedBalance } } = await this.gatewayService.get(`address/${this.apiConfigService.getAuctionContractAddress()}`);
+    let [ activeStake ] = await this.vmQueryService.vmQuery(
+      this.apiConfigService.getDelegationContractAddress(),
+      'getTotalActiveStake',
+    );
+    activeStake = this.numberDecode(activeStake);
+
+    let elrondConfig = {
```
bcosma, line 181: Shall we get this from an external config file? I assume they won't change too often (probably never) but you never know:))

bcosma (approved):

PR #101, opened 2021-08-29
Add denomination float

mihaieremia:
Added a new denominationFloat method which is using the old format function from express-api.


tanghel (changes requested):
- please re-open this PR against the development branch
- also, please use the Balance.toDenominated function of the erdjs framework for the computation
- last but not least, would be great to have some unit tests in the tests/helpers.spec.ts for this

Thank you! 💪

tanghel (dismissed):

AlexandruPislariu (dismissed):

src/endpoints/network/entities/networkConfig.ts
```diff
@@ -0,0 +1,10 @@
+import { ApiProperty } from '@nestjs/swagger';
+
+export class NetworkConfig {
```
AlexandruPislariu, line 3: Please change filename to `'network.config.ts'` to keep same structure.
Thank you!

AlexandruPislariu (approved):

tanghel (approved):

PR #103, opened 2021-08-30
validate node properties before applying search filter

AlexandruPislariu:


constantintovisi (dismissed):

src/endpoints/nodes/node.service.ts
```diff
@@ -94,9 +94,20 @@ export class NodeService {
 
     let filteredNodes = allNodes.filter(node => {
       if (query.search !== undefined) {
-        const nodeMatches = node.bls.toLowerCase().includes(query.search.toLowerCase());
```
tanghel, line 97: `const nodeMatches = node.bls && node.bls.toLowerCase().includes(query.search.toLowerCase());`
I think that should have been enough, since the same way are checked name & version beneath

tanghel (approved):

constantintovisi (approved):

PR #112, opened 2021-09-01
refactored calls to _doc to _search?q=_id format

tanghel:


src/common/bls.service.ts
```diff
@@ -22,28 +22,20 @@ export class BlsService {
...
-    const {
-      data: {
-        _source: { publicKeys },
-      },
-    } = await this.elasticService.get(url);
+    const url = `${this.url}/validators/_search?q=_id:${key}`;
   
+    let result = await this.elasticService.get(url);
+
+    let publicKeys = result.data.hits.hits[0]._source.publicKeys;
```
ccorcoveanu, line 29: I think we should at least check that `result.data.hits.hits > 0`

AlexandruPislariu (approved):

src/common/bls.service.ts
```diff
@@ -22,28 +22,20 @@ export class BlsService {
...
+
     this.publicKeysCache[key] = publicKeys;
   
     return publicKeys;
   };
 
   async getBlsIndex(bls: string, shardId: number, epoch: number): Promise<number | boolean> {
```
ccorcoveanu, line 36: Not related with this PR, but this function can have unexpected results because it can return `index ==0 0`. You always have to check `await getBlsIndex !== false` and it's prone to mistakes. We have one way of checking it the right way and a lot of ways to check this the wrong way :)

src/common/elastic.service.ts
```diff
@@ -38,8 +38,10 @@ export class ElasticService {
   };
 
   async getItem(collection: string, key: string, identifier: string) {
-    const url = `${this.url}/${collection}/_doc/${identifier}`;
-    const { data: document } = await this.get(url);
+    const url = `${this.url}/${collection}/_search?q=_id:${identifier}`;
+    let result = await this.get(url);
+
+    let document = result.data.hits.hits[0];
```
ccorcoveanu, line 44: Same as above, check for the length before accessing an empty array?

ccorcoveanu (changes requested):

src/common/bls.service.ts
tanghel: makes sense, done in both places where the search for an individual item is performed

src/common/bls.service.ts
tanghel: totally makes sense. the fix makes actually the code much more straightforward

AlexandruPislariu (approved):

ccorcoveanu (approved):

PR #115, opened 2021-09-02
website configuration for providers shouldn't be a must

AlexandruPislariu:


src/endpoints/providers/provider.service.ts
```diff
@@ -272,13 +272,16 @@ export class ProviderService {
...
+    if (response) {
       try {
-        const [name, website, identity] = response.map((base64) =>
-          Buffer.from(base64, 'base64').toString().trim().toLowerCase()
-        );
+        const [name, website, identity] = response.map((base64) => {
+          if(base64) {
+            return Buffer.from(base64, 'base64').toString().trim().toLowerCase();
+          }
+          return base64;
```
tanghel, line 281: you can return '' here for better control of the returned value

radumojic (approved):

PR #116, opened 2021-09-02
Added Transaction Scam Check logic

vladbucur1:


src/common/api.config.service.ts
```diff
@@ -252,4 +252,21 @@ export class ApiConfigService {
...
+      throw new Error('No extras api url present');
+    }
+
+    return extrasApiUrl;
+  }
+
+  getTransactionsScamCheck(): boolean {
+    var transactionsScamCheck = this.configService.get<boolean>('transactionsScamCheck');
+    if (transactionsScamCheck === undefined) {
+      throw new Error('No extras api url present');
```
tanghel, line 268: instead of throwing error, a default value can be returned so that this feature can only be disabled (enabled by default)

src/endpoints/transactions/scam-check/transaction-scam-check.service.ts
```diff
@@ -0,0 +1,81 @@
...
+      type,
+      info
+    };
+  }
+
+  private async loadScamTransactionResult(transaction: TransactionDetailed): Promise<ExtrasApiScamTransactionResult | null> {
+    const input = this.buildExtrasApiTransactionMinInfoDto(transaction);
+    return await this.cachingService.getOrSetCache(
+      `scam-info.${transaction.txHash}.${input.hasScResults}`,
+      () => this.extrasApiService.checkScamTransaction(input), 300, 300);
```
tanghel, line 65: instead of 300 you can use Constants.oneMinute() * 5 and leave the last parameter default

src/endpoints/transactions/scam-check/transaction-scam-check.service.ts
```diff
@@ -0,0 +1,81 @@
...
+    return {
+      type,
+      info
+    };
+  }
+
+  private async loadScamTransactionResult(transaction: TransactionDetailed): Promise<ExtrasApiScamTransactionResult | null> {
+    const input = this.buildExtrasApiTransactionMinInfoDto(transaction);
+    return await this.cachingService.getOrSetCache(
+      `scam-info.${transaction.txHash}.${input.hasScResults}`,
```
tanghel, line 64: is hasScResults relevant as the caching key or could it be left out altogether?

tanghel (changes requested):
please look at my comments in some of the files

tanghel (dismissed):

tanghel (approved):

PR #118, opened 2021-09-06
possibility to filter by receivershard if value is 0

tanghel:


AlexandruPislariu (approved):

src/endpoints/transactions/transaction.service.ts
```diff
@@ -51,11 +51,11 @@ export class TransactionService {
       queries.push(QueryType.Match('receiver', filter.receiver));
     }
 
-    if (filter.senderShard) {
+    if (filter.senderShard !== undefined) {
```
ccorcoveanu, line 54: Can the filter end up being `null` or empty string?

tanghel: the data type for this field is 'number | undefined', so at this point it should not have another value. the API endpoint is validating whether it should be a number or gives back undefined value

ccorcoveanu (approved):

PR #122, opened 2021-09-07
possibility to show owner on nfts/sfts list

AlexandruPislariu:


mihaieremia (approved):

src/endpoints/tokens/token.service.ts
```diff
@@ -164,32 +164,35 @@ export class TokenService {
...
-    if (nfts.length === 0) {
-      return undefined;
-    }
+    if (withOwner) {
+      let nftsDetailed: NftDetailed[] = [];
+      for (let nft of nfts) {
+        let nftDetailed: NftDetailed = ApiUtils.mergeObjects(new NftDetailed(), nft);
 
-    let nft: NftDetailed = ApiUtils.mergeObjects(new NftDetailed(), nfts[0]);
+        nftDetailed = await this.getTokenDistribution(nft.identifier, nftDetailed);
```
tanghel, line 177: here we need to perform a single query in the elasticsearch to fetch owner so that we don't perform 25 ES queries for a list of 25 items, and only do a single query in ES. This will be applied only in the case of nfts. in the case of sfts, it's too complex and should not be done.

tanghel (approved):

PR #133, opened 2021-09-15
add data url for statistics

AlexandruPislariu:


tanghel:
These settings are optional and should not be present in the standard config files

PR #137, opened 2021-09-17
heartbeatstatus optimizations

tanghel:


src/crons/cache.warmer.service.ts
```diff
@@ -102,6 +104,14 @@ export class CacheWarmerService {
     }, true);
   }
 
+  @Cron('* * * * *')
```
claudiulataretu, line 107: use cron expressions instead ```CronExpression.EVERY_MINUTE

claudiulataretu (approved):

AlexandruPislariu (approved):

PR #153, opened 2021-09-28
Transactions possibility to search by hash

AlexandruPislariu:


src/endpoints/transactions/transaction.service.ts
```diff
@@ -128,6 +128,27 @@ export class TransactionService {
...
     let transactions: Transaction[] = [];
 
+    if (filter.hashes) {
+      const txHashes: string[] = filter.hashes.split(',');
+      const elasticHashes = elasticTransactions.map(({txHash}) => txHash);
+      const missingHashes: string[] = txHashes.findMissingElements(elasticHashes);
+      let gatewayTransactions = await Promise.all(
+        missingHashes
+          .map(async (txHash) => {
+            const gatewayTransaction = await this.tryGetTransactionFromGateway(txHash);
```
tanghel, line 138: I would add here a parameter so that inside the tryGetTransactionFromGateway the tryGetTransactionFromElasticBySenderAndNonce function is not called

src/endpoints/transactions/transaction.service.ts
```diff
@@ -128,6 +128,27 @@ export class TransactionService {
 
     let transactions: Transaction[] = [];
 
+    if (filter.hashes) {
+      const txHashes: string[] = filter.hashes.split(',');
+      const elasticHashes = elasticTransactions.map(({txHash}) => txHash);
+      const missingHashes: string[] = txHashes.findMissingElements(elasticHashes);
+      let gatewayTransactions = await Promise.all(
+        missingHashes
+          .map(async (txHash) => {
```
tanghel, line 137: I would extract this in a function so that we only have something like await Promise.all(missingHashes.map(hash => this.tryGetTransactionFromGatewayForList(hash))). it would be a bit easier to read

tanghel (approved):

PR #168, opened 2021-10-03
Fix - avoid using default thumbnail if there is a custom one

andrei-x1:
* Avoid overriding `thumbnailUrl` by a default value for a custom thumbnail
* Removed redundant comma escape in regex

### DIFF
--- a/src/endpoints/tokens/token.service.ts
+++ b/src/endpoints/tokens/token.service.ts
@@ -280,7 +280,7 @@ export class TokenService {
 
       if (isCustomThumbnail === true) {
         nft.thumbnailUrl = `${this.apiConfigService.getMediaUrl()}/nfts/thumbnail/custom/${nft.identifier}`;
-      } if (isStandardThumbnail === true) {
+      } else if (isStandardThumbnail === true) {
         nft.thumbnailUrl = `${this.apiConfigService.getMediaUrl()}/nfts/thumbnail/standard/${nft.identifier}`;
       } else if (nft.metadata && nft.metadata.fileType) {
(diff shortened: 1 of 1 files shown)


src/endpoints/tokens/token.service.ts
```diff
@@ -280,7 +280,7 @@ export class TokenService {
 
       if (isCustomThumbnail === true) {
         nft.thumbnailUrl = `${this.apiConfigService.getMediaUrl()}/nfts/thumbnail/custom/${nft.identifier}`;
-      } if (isStandardThumbnail === true) {
+      } else if (isStandardThumbnail === true) {
```
tanghel, line 283: the custom thumbnail must have precedence, and in this situation, the custom thumbnail url will never be available, since the standard thumbnail will be taken with priority

tanghel (commented):
There is an issue with the if/else if logic in the thumbnails section. Anyway, this functionality is due to a refactoring, so that we have a standard url with some custom logic that will redirect to the custom/standard/default

src/endpoints/tokens/token.service.ts
andrei-x1: Adding the `else` keyword here assures that you say:
```custom > standard > defaults```

Without it, the URL containing "/custom" will never be used because it will be overridden in one of the following 3 lines starting with `nft.thumbnailUrl = ...` .

tanghel:
The logic here has changed, we don't have the concept of custom/standard/default thumbnails anymore

andrei-x1:
> 
> 
> The logic here has changed, we don't have the concept of custom/standard/default thumbnails anymore

@tanghel , ok. :)

PR #169, opened 2021-10-04
possibility to show SFT supply

AlexandruPislariu:


src/endpoints/tokens/token.service.ts
```diff
@@ -174,25 +174,43 @@ export class TokenService {
...
-            .toString();
+          .map((x: any) => BigInt(x.balance))
+          .reduce((previous: BigInt, current: BigInt) => previous.valueOf() + current.valueOf(), BigInt(0))
+          .toString();
+        }
+
+        if (withSupply) {
+          const { supply } = await this.cachingService.getOrSetCache(
+            `supply:${nft.identifier}`,
+            async () => await this.gatewayService.get(`network/esdt/supply/${nft.identifier}`),
```
tanghel, line 210: here it would be nice to have a function in the service to return the supply, so that we store the value directly decomposed in redis as well

tanghel (approved):

PR #174, opened 2021-10-06
define elastic utils and boilerplate

AlexandruPislariu:


src/common/elastic.service.ts
```diff
@@ -97,27 +97,23 @@ export class ElasticService {
...
-    const elasticQueryAdapter: ElasticQuery = new ElasticQuery();
-    elasticQueryAdapter.condition.should = identifiers.map((identifier) => QueryType.Match('identifier', identifier, QueryOperator.AND));
-    
-    let pagination = new ElasticPagination();
-    pagination.size = 10000;
-    elasticQueryAdapter.pagination = pagination;
+    const queries = identifiers.map((identifier) => QueryType.Match('identifier', identifier, QueryOperator.AND));
 
-    const elasticQuery = buildElasticQuery(elasticQueryAdapter);
+    const elasticQueryAdapter = ElasticUtils.boilerplate(QueryConditionOptions.should, queries, { from: 0, size: 10000 });
```
tanghel, line 115: instead of the boilerplate function in the ElasticUtils, you can have a static function create inside the ElasticQuery class. you can then use the builderPattern (https://en.wikipedia.org/wiki/Builder_pattern) as follows:
`
const elasticQuery = ElasticQuery.create()
   .withQueryShould(queries)
   .withPagination({ from: 0, size: 1000 });
`

src/common/elastic.service.ts
```diff
@@ -97,27 +97,23 @@ export class ElasticService {
...
+    const elasticQuery = ElasticUtils.buildElasticIndexerQuery(elasticQueryAdapter);
```
tanghel, line 116: this is also boilerplate. The getDocuments function below should accept an ElasticQuery parameter directly, and if it's ever used directly as an object, maybe create a `getDocumentsRaw` function to be lower-level

tanghel (approved):

PR #175, opened 2021-10-07
possibility to fast warm cron jobs

AlexandruPislariu:


src/common/api.config.service.ts
```diff
@@ -180,6 +180,15 @@ export class ApiConfigService {
     return isCronActive;
   }
 
+  getIsFastWarmerCronActive(): boolean {
+    let isCronActive = this.configService.get<boolean>('cron.fastWarm');
```
tanghel, line 184: instead of returning an error, let's fallback to false directly, so that we don't force config changes on all platforms

PR #235, opened 2021-10-27
Elrond API Extension

gabrielmatei:


src/utils/circular-queue-provider.ts
```diff
@@ -0,0 +1,96 @@
+export class Observer {
```
AlexandruPislariu, line 1: Maybe for this class should be created another file, to keep the same structure.

src/utils/circular-queue-provider.ts
```diff
@@ -0,0 +1,96 @@
...
+      for (const shard of sortedShards) {
+        const currentNodes = nodesMap.get(shard) ?? [];
+        const index = counterMap.get(shard) ?? 0;
+        if (index >= currentNodes.length) {
+          continue;
+        }
+
+        allNodes.push(currentNodes[index]);
+
+        step++;
```
AlexandruPislariu, line 82: `step` should be incremented inside for declaration

gabrielmatei: There are some cases in which we do not have to increment step

PR #281, opened 2021-11-09
Get num shards dynamically

AlexandruPislariu:


tanghel (commented):
We can implement getShardsRaw / getShards in the ShardService, and we can create similarily to the GenesisTimestampInterface an indirect way of providing the number of shards anywhere

tanghel (approved):

PR #315, opened 2021-11-19
handle expected errors

tanghel:


src/endpoints/esdt/esdt.service.ts
```diff
@@ -45,6 +45,8 @@ export class EsdtService {
       .withCondition(QueryConditionOptions.must, [ QueryType.Match('address', address) ])
       .withPagination({ from: 0, size: 10000 });
 
+    console.log('get from elastic');
```
AlexandruPislariu, line 48: This log can be removed

AlexandruPislariu (approved):

PR #405, opened 2022-01-13
Elastic profile get item

tanghel:
## Type
- [x] Bug
- [x] Feature  
- [ ] Performance improvement

## Problem setting
- Elastic search metrics are not reported when individual items are requested
  
## Proposed Changes
- Add metric for item request
- Remove took metric since it's not used any more
- Added elastic metric type to distinguish between list/item/count requests

## How to test
- perform various requests on endpoints that fetch data from elastic, such as /transactions or /blocks as list, item and count
- request metrics using http://localhost:4001/metrics and watch the correct buckets go up

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,9 +6,14 @@
 ## Problem setting
 - 
 - 
--
+- 
   
 ## Proposed Changes
--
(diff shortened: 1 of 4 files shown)


AlexandruPislariu (approved):

CatalinFaurPaul:
Functional test: work as expected

PR #483, opened 2022-02-22
filter nfts by collections

AlexandruPislariu:
## Type
- [x] Bug
- [ ] Feature  
- [ ] Performance improvement

## Problem setting
- Filter by collections array for address does not apply
  
## Proposed Changes
- Check if filter has collections property and build the elastic query accordingly


CatalinFaurPaul (approved):

src/endpoints/nfts/nft.service.ts
```diff
@@ -66,63 +66,71 @@ export class NftService {
...
+      mustQueries.push(QueryType.Match('identifier', identifier, QueryOperator.AND));
     }
 
     if (filter.collection !== undefined && filter.collection !== '') {
-      queries.push(QueryType.Match('token', filter.collection, QueryOperator.AND));
+      mustQueries.push(QueryType.Match('token', filter.collection, QueryOperator.AND));
+    }
+
+    if (filter.collections !== undefined && filter.collections.length !== 0) {
+      for (const collection of filter.collections) {
```
tanghel, line 94: you can use a similar approach to how identifiers are handled and then you wouldn't need shouldQueries any more. It will also be easier to extend in the future

tanghel (approved):

PR #509, opened 2022-03-03
refactoring integration test and added new test scenarios

CatalinFaurPaul:
## Type
- [ ] Bug
- [ ] Feature  
- [x] Integration Tests Improvement

### DIFF
--- a/src/test/integration/nodes.e2e-spec.ts
+++ b/src/test/integration/nodes.e2e-spec.ts
@@ -388,7 +388,7 @@ describe('Node Service', () => {
       const filter = new NodeFilter();
       filter.online = true;
       const count = await nodeService.getNodeCount(filter);
-      console.log(count);
+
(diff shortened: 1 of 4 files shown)


AlexandruPislariu (approved):

src/test/integration/nodes.e2e-spec.ts
```diff
@@ -276,7 +316,88 @@ describe('Node Service', () => {
   describe('Get Node Count', () => {
     it('should return node count', async () => {
       const count = await nodeService.getNodeCount(new NodeFilter());
-      expect(typeof count).toBe('number');
+      expect(typeof count).toStrictEqual('number');
```
AlexandruPislariu, line 319: I think it would be great to test that count is a real value, not only that it is of type number.
Maybe this type of scenarios can be performed with a large mocked list of nodes

tanghel (approved):

PR #548, opened 2022-03-18
relayedTx action support

tanghel:
## Type
- [ ] Bug
- [x] Feature  
- [ ] Performance improvement

## Proposed Changes
- support for decoding relayed transactions

## How to test (testnet)
- `/transactions/46408d0d3a90dc3e6a331ab757fd2868af1f1a4bcf84a1dcb798f09931a6bc6b` should return action data (esdtNft transfer buyChest)


src/endpoints/transactions/transaction-action/transaction.action.service.ts
```diff
@@ -95,6 +95,19 @@ export class TransactionActionService {
         metadata.functionName = dataComponents[0];
         metadata.functionArgs = args;
       }
+
+      if (metadata.functionName === 'relayedTx' && metadata.functionArgs.length === 1) {
```
miiu96, line 99: add support also for `relayedTxV2`

AlexandruPislariu (dismissed):

src/endpoints/transactions/transaction-action/transaction.action.service.ts
miiu96: also how you will treat the case you inside the `innerTx` data field contains another `relayedTx@hexEncodedTx`

tanghel: because it's a recursive call, there can be as many inner layers as needed

miiu96: Yes but you have to stop at the first inner tx because after that transaction will fail and the protocol will not execute more than one inner tx

CatalinFaurPaul (approved):

miiu96 (approved):

PR #562, opened 2022-03-30
Events processor

gabrielmatei:
## Type
- [ ] Bug
- [x] Feature  
- [ ] Refactoring
- [ ] Performance improvement

## Proposed Changes
- Process an NFT when `ESDTNFTCreate` event is received


src/common/rabbitmq/entities/nft/nft.topics.ts
```diff
@@ -0,0 +1,24 @@
+import { BinaryUtils } from "src/utils/binary.utils";
+import { NumberUtils } from "src/utils/number.utils";
+
+export class NftTopics {
+  identifier: string = '';
+  collection: string = '';
+  nonce: number = 0;
+
+  constructor(rawTopics: string[]) {
+    this.collection = BinaryUtils.base64Decode(rawTopics[0]);
```
AlexandruPislariu, line 10: Can you surround the decoding part with try catch and check if rawTopics[0] & rawTopics[1] actually exists?
So that if an error occurs you can log something like this 'Unexpected error when decoding NFT topics from raw topics'

src/common/rabbitmq/rabbitmq.nft.handler.service.ts
```diff
@@ -0,0 +1,28 @@
...
+export class RabbitMqNftHandlerService {
+
+  constructor(
+    private readonly nftWorkerService: NftWorkerService,
+    private readonly nftService: NftService,
+  ) { }
+
+  public async handleNftCreateEvent(event: NftCreateEvent): Promise<void> {
+    const identifier = event.getTopics()?.identifier;
+
```
AlexandruPislariu, line 17: I think here you should log that an ESDTNFTCreate was found 'Detected ESDTNFTCreate event for NFT with identifier'

src/common/rabbitmq/rabbitmq.nft.handler.service.ts
```diff
@@ -0,0 +1,28 @@
...
+    const nft = await this.nftService.getSingleNft(identifier);
+    if (!nft) {
```
AlexandruPislariu, line 19: I think here also a log is needed 'ESDTNFTCreate event: could not fetch nft details for NFT with identifier'

src/common/rabbitmq/rabbitmq.nft.handler.service.ts
```diff
@@ -0,0 +1,28 @@
...
+      return;
+    }
+
+    const needsProcessing = await this.nftWorkerService.needsProcessing(nft, new ProcessNftSettings());
```
AlexandruPislariu, line 23: Also, the nft processing should be surrounded with try catch and log if an error occurs related to processing 'NFTProcess: Unexpected error when processing NFT with identifier'

AlexandruPislariu (dismissed):

PR #595, opened 2022-04-27
GraphQL query adapter

AlexandruPislariu:
## Type
- [ ] Bug
- [ ] Feature  
- [ ] Refactoring
- [ ] Performance improvement

## Problem setting
- No support for graphql DSL
  
## Proposed Changes
- Add a service that acts as a wrapper for all graphql queries

### DIFF
--- a/src/common/api-config/api.config.service.ts
+++ b/src/common/api-config/api.config.service.ts
@@ -587,4 +587,17 @@ export class ApiConfigService {
   getNftProcessMaxRetries(): number {
     return this.configService.get<number>('nftProcess.maxRetries') ?? 3;
   }
+
+  getMicroServiceUrl(): string | undefined {
+    return this.configService.get<string>('transaction-action.mex.microServiceUrl') ?? this.configService.get<string>('plugins.transaction-action.mex.microServiceUrl');
+  }
+
+  getMicroServiceUrlMandatory(): string {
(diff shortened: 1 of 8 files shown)


src/common/graphql/graphql.service.ts
```diff
@@ -0,0 +1,36 @@
...
+export class GraphQlService {
+  private readonly logger: Logger;
+  private readonly MICROSERVICE_URL: string;
+  private readonly graphqlClient: GraphQLClient;
+
+  constructor(
+    private readonly apiConfigService: ApiConfigService
+  ) {
+    this.logger = new Logger(GraphQlService.name);
+    this.MICROSERVICE_URL = this.apiConfigService.getMicroServiceUrlMandatory();
```
tanghel, line 15: since this is only used in the constructor, you can use a local variable for microservice url

src/common/api-config/api.config.service.ts
```diff
@@ -587,4 +587,17 @@ export class ApiConfigService {
   getNftProcessMaxRetries(): number {
     return this.configService.get<number>('nftProcess.maxRetries') ?? 3;
   }
+
+  getMicroServiceUrl(): string | undefined {
```
tanghel, line 591: getMicroServiceUrl is too generic.
maybe a better name would be `getMaiarExchangeUrl`

tanghel (approved):

PR #607, opened 2022-05-04
Token multiple assets

tanghel:
## Type
- [ ] Bug
- [x] Feature  
- [ ] Refactoring
- [ ] Performance improvement

## Proposed Changes
- Support for declaring `extraTokens` in token assets, with the effect of calculating total number of token accounts / transactions taking into consideration other similar tokens (LP tokens, Farm tokens, Metastaking, etc.)

## How to test (mainnet)
- `/tokens/RIDE-7d18e9` should contain > 70k accounts, > 1mil transactions
- `/tokens/RIDE-7d18e9/supply` should also contain locked accounts information

### DIFF
--- a/src/endpoints/tokens/token.service.ts
+++ b/src/endpoints/tokens/token.service.ts
@@ -114,30 +114,34 @@ export class TokenService {
     }
 
     if (filter.sort) {
-      const order = filter.order ?? SortOrder.desc;
+      tokens = this.sortTokens(tokens, filter.sort, filter.order ?? SortOrder.desc);
+    }
 
-      let criteria: (token: Token) => number;
+    return tokens;
+  }
 
-      switch (filter.sort) {
-        case TokenSort.accounts:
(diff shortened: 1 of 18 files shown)


src/endpoints/tokens/token.service.ts
```diff
@@ -173,6 +113,33 @@ export class TokenService {
       tokens = tokens.filter(token => identifierArray.includes(token.identifier.toLowerCase()));
     }
 
+    if (filter.sort) {
+      const order = filter.order ?? SortOrder.desc;
+
+      let criteria: (token: Token) => number;
+
```
danielailie, line 120: I would extract each switch into separate method

danielailie (approved):

src/endpoints/tokens/token.service.ts
tanghel: extracted into a single, unified function for sorting

PR #614, opened 2022-05-06
Calculate token accounts and transactions for all tokens

tanghel:
## Type
- [x] Bug
- [ ] Feature  
- [ ] Refactoring
- [ ] Performance improvement

## Proposed Changes
- Calculate token accounts & transactions for all tokens, not only those that have assets

## How to test (mainnet)
- After waiting a few minutes so that Esdt tokens invalidations cache warmer is finished: `/tokens/UPARK-982dd6` should contain > 1000 accounts & transactions


CatalinFaurPaul (commented):
Test passed

PR #652, opened 2022-05-25
Initialize PublicApp only if the public API flag is set

tanghel:
## Type
- [ ] Bug
- [x] Feature  
- [ ] Refactoring
- [ ] Performance improvement

## Proposed Changes
- Initialize PublicApp only if the public API flag is set

## How to test
- Modules dependent from the public app should not be initialized unless the public api flag is set


src/main.ts
```diff
@@ -27,93 +27,91 @@ import { PaginationInterceptor } from './interceptors/pagination.interceptor';
...
-      new CachingInterceptor(
-        cachingService,
-        httpAdapterHostService,
-        metricsService,
-        protocolService,
-      ),
+  if (apiConfigService.getIsPublicApiActive()) {
+    const publicApp = await NestFactory.create<NestExpressApplication>(PublicAppModule);
+    publicApp.use(bodyParser.json({ limit: '1mb' }));
+    publicApp.use(requestIp.mw());
```
danielailie, line 47: maybe here you can extract a method configurePublicApi() or getPublicApi

src/main.ts
```diff
@@ -27,93 +27,91 @@ import { PaginationInterceptor } from './interceptors/pagination.interceptor';
...
+    publicApp.enableCors();
+    publicApp.useLogger(publicApp.get(WINSTON_MODULE_NEST_PROVIDER));
+    publicApp.disable('etag');
+    publicApp.disable('x-powered-by');
+
```
danielailie, line 52: move this closer to their usage and define them only if need it

claudiulataretu (dismissed):

danielailie (approved):

gabrielmatei (dismissed):

CatalinFaurPaul (dismissed):

src/main.ts
tanghel: indeed

src/main.ts
tanghel: these are part of the default configuration, have no usage in main.ts

tanghel: fixed

danielailie (approved):

PR #676, opened 2022-06-13
Empty types in parameters in generated swagger schema json file

JotaStar:
## Type
- [X] Bug
- [ ] Feature
- [ ] Refactoring
- [ ] Performance improvement
- [ ] Integration test

## Problem setting

Some parameter types are defined as nullable in the form "type | undefined" instead of "type?"

In this format, swagger does not recognize the type in the json schema file it generates in the /api-json path, leaving the "schema" node empty

In addition, the optional pipes, by returning undefined, also mask the data type.

## Proposed Changes

Replace the definition of nullable types of the form "type | undefined" to "type?" and return string instead undefined in optional pipes that doesn't affect to the pipe.

## How to test

Get the json schema file from swagger in the path /api-json and check, for example, in "blocks/count" the parameters "shard", "proposer", etc...


src/utils/pipes/parse.array.pipe.ts
```diff
@@ -11,7 +11,7 @@ export class ParseArrayPipe implements PipeTransform<string | undefined, Promise
   transform(value: string | undefined, _: ArgumentMetadata): Promise<string[] | undefined> {
     return new Promise(resolve => {
       if (value === undefined || value === '') {
-        return resolve(undefined);
+        return resolve([""]);
```
tanghel, line 14: this may cause unintended consequences in endpoints that accept arrays. There are checks for undefined where the filters are applied which might break the existing functionality

src/utils/pipes/parse.optional.enum.pipe.ts
```diff
@@ -6,7 +6,7 @@ export class ParseOptionalEnumPipe<T extends { [name: string]: any }> implements
   transform(value: string | undefined, _: ArgumentMetadata): Promise<string | undefined> {
     return new Promise(resolve => {
       if (value === undefined || value === '') {
-        return resolve(undefined);
+        return resolve('');
```
tanghel, line 9: Same as the above comment. There are service-level checks for undefined as an explicit value

tanghel:
Hello! Thank you for taking your time to create this helpful PR!

Before we can merge this, please re-target this PR on the `development` branch, because that's where we bring new features, refactorings, etc.

Also, the changes for parse.array.pipe.ts and parse.optional.enum.pipe.ts will break existing functionality

PR #689, opened 2022-06-17
Add 404 status code for collections endpoint

CatalinFaurPaul:
## Type
- [ ] Bug
- [ ] Feature  
- [x] Refactoring
- [ ] Performance improvement
- [ ] Integration test

## Problem setting
- There was no collection not found condition for endpoints:
 - /collections/:collection/nfts/count
 - /collections/:collection/nfts  
  
## Proposed Changes
- Add condition to verify if collection exist

## How to test
-  /collections/Test/nfts -> should return 404 instead of []
- /collections/Test/nfts/count -> should return 404 instead of 0


tanghel (changes requested):
I would suggest here to create a function called `isCollection`, similar how it is done for tokens

tanghel (approved):

PR #694, opened 2022-06-21
Create endpoint for triggering NFT media/metadata reprocessing

gabrielmatei:
## Problem setting
- The collection owner cand trigger NFT media/metadata reprocessing

## Proposed Changes
- Add and endpoint that triggers the NFT media/metadata reprocessing

## How to test
- `POST /nfts/process` called by the collection owner triggers reprocessing


src/endpoints/process-nfts/process.nfts.service.ts
```diff
@@ -44,4 +90,48 @@ export class ProcessNftsService {
...
+      throw new HttpException(`The owner's address could not be found`, HttpStatus.BAD_REQUEST);
+    }
+
+    let collectionOwner = nftCollection.owner;
+
+    let currentDepth = 0;
+    while (AddressUtils.isSmartContractAddress(collectionOwner) && currentDepth < ProcessNftsService.MAX_DEPTH) {
+      const {
+        account: { ownerAddress },
+      } = await this.gatewayService.get(`address/${collectionOwner}`, GatewayComponentRequest.addressDetails);
```
tanghel, line 123: You can request account details from the accountService, for more integrity

src/endpoints/process-nfts/process.nfts.service.ts
```diff
@@ -44,4 +90,48 @@ export class ProcessNftsService {
 
     return await this.nftWorkerService.addProcessNftQueueJob(nft, settings);
   }
+
+  private async checkCollectionOwner(address: string, collection?: string, identifier?: string): Promise<void> {
+    if (identifier) {
+      const nft = await this.nftService.getSingleNft(identifier);
+      if (!nft) {
+        throw new HttpException('Provide a valid identifier or a collection to generate thumbnails for', HttpStatus.BAD_REQUEST);
```
tanghel, line 98: at this level you should throw normal errors, since theoretically the service layer is not dependent on http calls.
You can try/catch it in the controller and re-package the error there as BadRequest
Also, you can throw new BadRequestException which does the same but is a bit better integrated

src/interceptors/access.interceptor.ts
```diff
@@ -4,13 +4,13 @@ import { ApiConfigService } from "src/common/api-config/api.config.service";
 import { HealthCheckController } from "src/endpoints/health-check/health.check.controller";
 
 @Injectable()
-export class JwtAuthenticateGuard implements CanActivate {
+export class JwtAuthenticateGlobalGuard implements CanActivate {
```
tanghel, line 7: Since this is a guard and not an interceptor, you can move it to the utils/guards folder and rename the file accordingly

tanghel (dismissed):

PR #723, opened 2022-07-11
Upload nft asset to S3

AlexandruPislariu:
## Problem setting
- NFT original assets retrieved from IPFS or other slow storages
  
## Proposed Changes
- Process NFT original asset (download from IPFS & upload to S3)
- Add a cronjob to process old NFT collections
- Add possibility to request forceUploadAsset from /nfts/process endpoint [public & private(owner only)]

## How to test
- Find one nft for which original asset isn't uploaded to S3 (https://media.elrond.com)
- Use /nfts/process with forceUploadAsset=true 
- Check if original asset from S3 is the same with the one from original storage

### DIFF
--- a/src/crons/nft/nft.cron.service.ts
+++ b/src/crons/nft/nft.cron.service.ts
@@ -1,11 +1,14 @@
-import { Constants, Locker } from "@elrondnetwork/erdnest";
+import { CachingService, Constants, Locker } from "@elrondnetwork/erdnest";
 import { Injectable, Logger } from "@nestjs/common";
 import { Cron, CronExpression } from "@nestjs/schedule";
 import { ApiConfigService } from "src/common/api-config/api.config.service";
+import { QueryPagination } from "src/common/entities/query.pagination";
 import { Nft } from "src/endpoints/nfts/entities/nft";
+import { NftFilter } from "src/endpoints/nfts/entities/nft.filter";
 import { NftService } from "src/endpoints/nfts/nft.service";
 import { ProcessNftSettings } from "src/endpoints/process-nfts/entities/process.nft.settings";
 import { NftWorkerService } from "src/queue.worker/nft.worker/nft.worker.service";
+import { CacheInfo } from "src/utils/cache.info";
 
 @Injectable()
 export class NftCronService {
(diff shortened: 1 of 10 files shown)


gabrielmatei (approved):

src/queue.worker/nft.worker/queue/nft.queue.controller.ts
```diff
@@ -91,6 +93,10 @@ export class NftQueueController {
         await Promise.all(nft.media.map((media: any) => this.generateThumbnail(nft, media, settings.forceRefreshThumbnail)));
       }
 
+      if (nft.media) {
+        await Promise.all(nft.media.map((media: NftMedia) => this.nftAssetService.uploadAsset(nft.identifier, media.originalUrl, media.fileType)));
```
tanghel, line 97: I would implement here a verification whether asset was already uploaded or not

src/crons/nft/nft.cron.service.ts
```diff
@@ -37,6 +41,27 @@ export class NftCronService {
...
+      return;
+    }
+
+    let lastProcessedTimestamp = await this.cachingService.getCache<number>(CacheInfo.LastProcessedTimestamp.key) ?? Math.floor(Date.now() / 1000);
+    this.logger.log(`Last processed timestamp until now: ${lastProcessedTimestamp} (${new Date(lastProcessedTimestamp * 1000)})`);
+
+    await Locker.lock('Process NFTs whitelisted', async () => {
+      const nfts = await this.nftService.getNfts(new QueryPagination({ from: 0, size: 2500 }), new NftFilter({ isWhitelistedStorage: true, hasUris: true, before: lastProcessedTimestamp }));
+
+      await Promise.all(nfts.map((nft: Nft) => this.nftWorkerService.addProcessNftQueueJob(nft, new ProcessNftSettings({ uploadAsset: true }))));
```
tanghel, line 56: I would make the check before triggering the rabbitmq, so that we don't overload rmq with messages

src/crons/nft/nft.cron.service.ts
```diff
@@ -64,9 +64,11 @@ export class NftCronService {
           continue;
         }
 
-        const isAssetUploaded = await this.nftAssetService.isAssetUploaded(nft.identifier, nft.media[0]);
-        if (!isAssetUploaded) {
-          needProccessNfts.push(nft);
+        for (const media of nft.media) {
+          const isAssetUploaded = await this.nftAssetService.isAssetUploaded(media);
+          if (!isAssetUploaded) {
+            needProccessNfts.push(nft);
```
AlexandruPislariu, line 70: If an NFT have more than one media files it will be processed more than one time

tanghel (approved):

PR #764, opened 2022-08-04
Update NFT metadata on refresh

AlexandruPislariu:
- No need to delete metadata from database since on set the value is updated
  
## Proposed Changes
- Just set metadata on refresh

## How to test
- Create a new NFT, check if metadata is returned
- Trigger a force refresh on metadata


tanghel (approved):

CatalinFaurPaul:
Test passed

PR #767, opened 2022-08-05
Nft media needs processing

AlexandruPislariu:
- Nft worker will never considerate that NFT needs processing if it doesn't have media
  
## Proposed Changes
- Check if NFT has default media, if it has then needs processing

## How to test
- Create an NFT and do not trigger process on NFT create
- Tigger NFT process manually or check using cron if processing is triggered


tanghel (approved):

CatalinFaurPaul:
Test passed

PR #774, opened 2022-08-10
get transaction from gateway only if not found in elastic

AlexandruPislariu:
- If Elastic is down the load is redirected to gateway
  
## Proposed Changes
- Fallback to gateway only if transaction is not found in elastic 

## How to test
- Check fallback if elastic is down
- Check if a transaction is returned if exists only on gateway


CatalinFaurPaul:
Test passed

PR #778, opened 2022-08-11
Services 482 mongodb transaction details

gabrielmatei:
- use database as a caching layer to store transaction details.

## Proposed Changes
- when a transaction is fetched, first look for it in database. If not found, fetch from ES, then store all details in database (if status <> pending), then return it
- introduce a “L0” cache in redis / memory with ttl of 10 minutes


src/endpoints/transactions/transaction.get.service.ts
```diff
@@ -211,4 +216,49 @@ export class TransactionGetService {
...
+
+    const txFromElastic = await this.tryGetTransactionFromElastic(txHash, fields);
+    if (txFromElastic) {
+      await this.cacheTransaction(txFromElastic);
+      return txFromElastic;
+    }
+
+    const txFromGateway = await this.tryGetTransactionFromGateway(txHash);
+    if (txFromGateway) {
+      await this.cacheTransaction(txFromGateway);
```
AlexandruPislariu, line 240: I think we should not cache a transaction from gateway, because if a transaction is not yet indexed in elastic we will always return the gateway version for it, which is incomplete.

docker-compose.yml
```diff
@@ -47,3 +47,10 @@ services:
     ports:
       - 5673:5672
       - 15673:15672
+
+  mongodb:
+    image: mongo:latest
+    environment:
+      - MONGODB_DATABASE="api"
+    ports:
+      - 27017:27017
```
johnykes, line 56: new line

PR #840, opened 2022-10-06
 filter nfts by nonce before after

CatalinFaurPaul:
## Proposed Changes
-  add nonce filter

## How to test
REST API
/nfts:
- nfts?nonce=200 -> should return all NFTS/SFTS that contains given nonce
- nfts/count?nonce=300 -> should return count 516
- nfts/c?nonce=300 -> should return count 516 

/collections/:collection/nfts
- collections/ANGELSLOGO-4adf87/nfts?nonce=8 -> should return ANGELSLOGO-4adf87-08 NFT with nonce = 8
- collections/BACPASS-04e709/nfts?nonceBefore=10 -> should return all NFTs from specified collection in descending order

API GraphQL
- create nfts query and add nonce: 200 as input

### DIFF
--- a/src/test/integration/services/nft.e2e-spec.ts
+++ b/src/test/integration/services/nft.e2e-spec.ts
@@ -311,6 +311,17 @@ describe('Nft Service', () => {
         expect(result.timestamp).toBeGreaterThanOrEqual(1654630698);
       }
     });
+
+    it('should return 10 NFTS that contains nonce equal with 200', async () => {
+      const filter = new NftFilter();
+      filter.nonce = 200;
+
+      const results = await nftService.getNfts({ from: 0, size: 10 }, filter);
+
+      for (const result of results) {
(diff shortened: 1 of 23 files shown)


src/test/integration/services/nft.e2e-spec.ts
```diff
@@ -381,6 +392,14 @@ describe('Nft Service', () => {
       expect(count).toStrictEqual(10000);
     });
 
+    it("should return total number of NFTs that contains nonce equal with 200 ", async () => {
```
AlexandruPislariu, line 395: I think this test is not accurate enough, maybe extend it with before & after to know the exact count of nfts.
Also, a test scenario with nonce for collection nfts would work.

CatalinFaurPaul: Test was refactored with after timestamp filter

PR #842, opened 2022-10-10
 add herotag details in transactions

CatalinFaurPaul:
## Proposed Changes
- Add herotag support in transaction list & details

## How to test(testnet)
- `/transactions?withUsername=true`, `/transfers?withUsername=true` and all related endpoints should return the herotag as assets wherever an address is present
- on all transaction & transfer endpoints, if `withUsername` flag is activated, a maximum number of 50 items should be allowed
- the same logic should also function correctly on the graphql endpoints

### DIFF
--- a/src/endpoints/transactions/entities/transactions.query.options.ts
+++ b/src/endpoints/transactions/entities/transactions.query.options.ts
@@ -12,7 +12,7 @@ export class TransactionQueryOptions {
   withScamInfo?: boolean;
   withUsername?: boolean;
 
-  static applyDefaults(size: number, options: TransactionQueryOptions): TransactionQueryOptions {
+  static applyDefaultOptions(size: number, options: TransactionQueryOptions): TransactionQueryOptions {
     if (size <= TransactionQueryOptions.SIZE_LIMIT) {
       options.withScamInfo = true;
       options.withUsername = true;
--- a/src/endpoints/transactions/entities/transactions.query.options.ts
+++ b/src/endpoints/transactions/entities/transactions.query.options.ts
@@ -10,6 +10,7 @@ export class TransactionQueryOptions {
   withLogs?: boolean = true;
   withScResultLogs?: boolean = true;
   withScamInfo?: boolean;
+  withHerotag?: boolean;
 
   //TODO: Remove this function when enforce is no longer needed
   static enforceScamInfoFlag(size: number, options: TransactionQueryOptions): TransactionQueryOptions {
--- a/src/endpoints/transactions/entities/transactions.query.options.ts
+++ b/src/endpoints/transactions/entities/transactions.query.options.ts
@@ -1,5 +1,8 @@
+import { BadRequestException } from "@nestjs/common";
+
 export class TransactionQueryOptions {
-  private static readonly SIZE_LIMIT: number = 50;
+  private static readonly SCAM_INFO_MAX_SIZE: number = 50;
+  private static readonly USERNAME_MAX_SIZE: number = 50;
 
   constructor(init?: Partial<TransactionQueryOptions>) {
     Object.assign(this, init);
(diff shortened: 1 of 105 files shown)


gabrielmatei (approved):

src/endpoints/transactions/transaction.service.ts
```diff
@@ -227,7 +312,7 @@ export class TransactionService {
     }
   }
 
-  async processTransactions(transactions: Transaction[], withScamInfo: boolean, assets?: Record<string, AccountAssets>): Promise<void> {
+  async processTransactions(transactions: Transaction[], withScamInfo: boolean, withUsername: boolean): Promise<void> {
```
dragos-rebegea, line 315: processTransactions can have as a parameter an `DefaultTransactionQueryOptions`, this class can be also.
This way it will be more clear that the first bool value is for ScamInfo and the 2nd one for Username. Also this way also the developer error will be removed, since he won`t be able anymore to mess up the order of the two adjacent parameters with the same type.

`TransactionQueryOptions` can also extend this class the default query options can be kept separate.

src/endpoints/transfers/transfer.service.ts
```diff
@@ -65,7 +62,7 @@ export class TransferService {
       transactions.push(transaction);
     }
 
-    await this.transactionService.processTransactions(transactions, pagination.size <= 100, assets);
+    await this.transactionService.processTransactions(transactions, pagination.size <= 100, pagination.size <= 50);
```
dragos-rebegea, line 65: the constants can be extracted

src/endpoints/usernames/username.service.ts
```diff
@@ -1,16 +1,36 @@
...
+    private readonly apiService: ApiService,
+    private readonly apiConfigService: ApiConfigService,
+    private readonly vmQueryService: VmQueryService
   ) { }
 
-  private async getUsernameAddress(username: string): Promise<string | null> {
+  async getUsernameForAddressRaw(address: string): Promise<string | null> {
+    try {
+      // eslint-disable-next-line require-await
+      const result = await this.apiService.get(`${this.apiConfigService.getMaiarIdUrl()}/users/api/v1/users/${address}`, undefined, async error => error?.response?.status === HttpStatus.FORBIDDEN);
```
dragos-rebegea, line 21: is the eslinter-disable required ?

src/endpoints/transactions/entities/transactions.query.options.ts
```diff
@@ -10,11 +10,12 @@ export class TransactionQueryOptions {
   withLogs?: boolean = true;
   withScResultLogs?: boolean = true;
   withScamInfo?: boolean;
+  withUsername?: boolean;
 
-  //TODO: Remove this function when enforce is no longer needed
-  static enforceScamInfoFlag(size: number, options: TransactionQueryOptions): TransactionQueryOptions {
+  static applyDefaults(size: number, options: TransactionQueryOptions): TransactionQueryOptions {
```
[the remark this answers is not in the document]
tanghel: fixed

src/endpoints/usernames/username.service.ts
tanghel: yes, because the callback is a Promise and we don't execute any await, it gives a warning
but in this particular situation, we know what we're doing 💪

src/endpoints/transfers/transfer.service.ts
tanghel: refactored vigorously, extracted in constants and taken them into account

src/endpoints/transactions/transaction.service.ts
tanghel: created an options object also in this situation

src/endpoints/transactions/entities/transactions.query.options.ts
```diff
@@ -1,5 +1,8 @@
+import { BadRequestException } from "@nestjs/common";
+
 export class TransactionQueryOptions {
-  private static readonly SIZE_LIMIT: number = 50;
+  private static readonly SCAM_INFO_MAX_SIZE: number = 50;
```
dragos-rebegea, line 4: this one previous value was 100. Is it the current value intended?

tanghel: yes, this is intentional

dragos-rebegea (approved):

PR #847, opened 2022-10-12
Service ext 1 transfer add sc results

CristianFodor:

I need the smart contracts results on account transfers list for getting the stake values.
  
## Proposed Changes
Add flag withScResult on account transfers. Later I'll propose a new way of adding this data directly on action informations.


tanghel:
Hello!
Maybe this needs further clarification.
The `/transfers` endpoint already contains all transactions + scResults affecting a given account.
Do we in this situation still need to fetch the scResults for every transaction?

PR #852, opened 2022-10-17
create collection roles endpoint

CatalinFaurPaul:
## Proposed Changes
-  Create an endpoint for returning collection roles

## How to test
-  collections/CITIZEN-375553/roles -> should return collection roles

```
    {
        "address": "erd1qqqqqqqqqqqqqpgq6ggn6j69c8x3808zt0rudnlh4hrqzf20sdpsdmmta5",
        "canCreate": true,
        "canBurn": false,
        "canAddQuantity": false,
        "canUpdateAttributes": false,
        "canAddUri": false,
        "canTransferRole": false,
        "roles": [
            "ESDTRoleNFTCreate"
        ]
    }
```

### DIFF
--- a/src/common/indexer/entities/collection.ts
+++ b/src/common/indexer/entities/collection.ts
@@ -7,4 +7,5 @@ export interface Collection {
   type: string;
   timestamp: number;
   ownersHistory: { address: string, timestamp: number }[];
+  roles: any
 }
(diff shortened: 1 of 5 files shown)


src/common/indexer/entities/collection.ts
```diff
@@ -7,4 +7,5 @@ export interface Collection {
   type: string;
   timestamp: number;
   ownersHistory: { address: string, timestamp: number }[];
+  roles: any
```
AlexandruPislariu, line 10: I think we can use the type 'CollectionRoles' instead of any.

PR #877, opened 2022-10-30
 type safe interactions with gateway

CatalinFaurPaul:
## Proposed Changes
- Instead of calling generic get/post methods on the gateway,  implement for every endpoint on the gateway a function with explicit parameters and return type

## How to test
- every call from gateway service should return correct data


gabrielmatei (dismissed):

gabrielmatei:
LGTM

tanghel (approved):

PR #878, opened 2022-10-31
Process collections

AlexandruPislariu:
- Missing the ability to process collections
  
## Proposed Changes
- Add processCollection method in PluginsService

## How to test
- Write a custom processCollection function, check if it is called whenever /accounts/:address/collections, /collections, /collections/:collection, /accounts/:address/collections/:collection endpoints are requested.


src/endpoints/collections/collection.module.ts
```diff
@@ -11,6 +12,7 @@ import { CollectionService } from "./collection.service";
     forwardRef(() => VmQueryModule),
     forwardRef(() => TokenModule),
     forwardRef(() => AssetsModule),
+    PluginModule,
```
CatalinFaurPaul, line 15: A circular dependency between modules. Use forwardRef() to avoid it.
 forwardRef(() => PluginModule)

AlexandruPislariu: Done

CatalinFaurPaul (dismissed):

tanghel (approved):

PR #889, opened 2022-11-10
decouple ApiMetricsService from application

florincoz:
- decouple api-metrics-service from the rest of the app
- remove some circular dependencies
  
## Proposed Changes
- add event based metrics trigger


CatalinFaurPaul:
@nestjs/event-emitter should be added in package.json

gabrielmatei (approved):

gabrielmatei:
we should wait for erdnest update

PR #890, opened 2022-11-10
Show transaction prices

gabrielmatei:
- we need to display the historical prices for transaction operations in the [Explorer](https://explorer.multiversx.com/)
  
## Proposed Changes
- use the plugin service to get the historical prices for ESDTs
- ESDT prices are supported only for the tokens listed on [xExchange](https://xexchange.com/) and [Jungle DEX](https://jungledex.com/)
## How to test
- `/transactions?function=claimRewards&withOperations=true&withEsdtPrices=true` - it should return `valueUsd` for each operation with supported tokens


gabrielmatei:
wait for plugins upgrade

tanghel (approved):

tanghel (approved):

PR #897, opened 2022-11-22
Unified Tokens & MetaESDTs

tanghel:
## Reasoning 
- Although structurally similar, an NFT of type `MetaESDT` has much more similarity to a payable token, rather than an NFT
- what is actually different between a fungible token and a MetaESDT is that every instance of it can carry certain metadata in the form of attributes
- this is how e.g. LKMEX unlock schedule is encoded or how farming reward positions are stored
  
## Proposed Changes
- in all token-related endpoints, offer the possibility to load both tokens of type `FungibleESDT` and `MetaESDT`
- in all nft-related endpoints, offer the possibility to not load tokens of type `MetaESDT`

## How to test (mainnet)
- `/tokens?includeMetaESDT=true` should return both tokens of type `FungibleESDT` and `MetaESDT`
- `/tokens?includeMetaESDT=true&type=MetaESDT` should return only tokens of type `MetaESDT`
- `/tokens/:token` should return token details even if type = `MetaESDT`
- `/tokens/:token/accounts` should also return data for token collections of type `MetaESDT` and will return token identifier & attributes extra compared to a fungible token
- `/tokens/:token/supply` should return supply information also for tokens of type `MetaESDT`. For now, all values are 0 but in the future, when the node saves collection supply, that information will be provided
- `/accounts/:account/tokens?includeMetaESDT=true` should return both tokens of type `FungibleESDT` and `MetaESDT` from the specified account
  - it should also contain the `attributes` field in case of `MetaESDT`
- `/accounts/:account/tokens?includeMetaESDT=true&type=MetaESDT` should return only tokens of type `MetaESDT` from the specified account
- `/accounts/:account/tokens/:token` should return account token details also if its type is `MetaESDT`
- `/collections?excludeMetaESDT=true` should return only collections of type `NonFungibleESDT` and `SemiFungibleESDT` (without the ones of type `MetaESDT`)
- `/accounts/:account/roles/collections?excludeMetaESDT=true` should return only collections of type `NonFungibleESDT` and `SemiFungibleESDT` (without the ones of type `MetaESDT`) from the specified account
- `/accounts/:account/collections?excludeMetaESDT=true` should return only collections of type `NonFungibleESDT` and `SemiFungibleESDT` (without the ones of type `MetaESDT`) from the specified account
- `/accounts/:account/roles/tokens?includeMetaESDT=true` should return both tokens of type `FungibleESDT` and `MetaESDT` from the specified account
- `/accounts/:account/nfts?excludeMetaESDT=true` should return only NFTs of type `NonFungibleESDT` and `SemiFungibleESDT` (without the ones of type `MetaESDT`) from the specified account


gabrielmatei:
`/tokens?withMetaESDT=true&type=MetaESDT should return only tokens of type FungibleESDT` - shouldn't it only return `MetaESDT` tokens?

tanghel:
> `/tokens?withMetaESDT=true&type=MetaESDT should return only tokens of type FungibleESDT` - shouldn't it only return `MetaESDT` tokens?

Indeed, there was a problem in the PR description. Should be good now

CatalinFaurPaul (approved):

CatalinFaurPaul (approved):

PR #908, opened 2022-12-12
Cached dapp config

tanghel:
## Problem setting
- Dapp config is pretty static and api calls are accessing the disk directly
  
## Proposed Changes
- Fetch dapp config only once and keep it in memory

## How to test
- Make sure the dapp config route does not access disk any more


CatalinFaurPaul:
Test Passed

PR #915, opened 2022-12-14
 add deployTxHash attribute

CatalinFaurPaul:
### Reasoning
 -  When contract address details are returned, deployedTxHash should be defined
## Proposed Changes
- Add deployedTxHash attribute
- Add account/:address/upgrades endpoint

## How to test
- accounts/erd1qqqqqqqqqqqqqpgqeel2kumf0r8ffyhth7pqdujjat9nx0862jpsg2pqaq -> deployedTxHash should be defined
- accounts/erd1qqqqqqqqqqqqqpgqeel2kumf0r8ffyhth7pqdujjat9nx0862jpsg2pqaq/upgrades -> should return contract upgrades details


src/endpoints/accounts/account.controller.ts
```diff
@@ -889,6 +890,17 @@ export class AccountController {
     return this.accountService.getAccountContractsCount(address);
   }
 
+  @Get("/accounts/:address/upgrades")
+  @ApiOperation({ summary: 'Account upgrades details', description: 'Returns all upgrades details for a specific contract address' })
+  @ApiOkResponse({ type: ContractUpgrades })
+  getContractUpgrades(
+    @Param('address', ParseAddressPipe) address: string,
+  ): Promise<ContractUpgrades | null> {
+    const upgrades = this.accountService.getContractUpgrades(address);
```
gabrielmatei, line 899: i think we should throw 404 error when `upgrades` is `null`

gabrielmatei (changes requested):

gabrielmatei (approved):

PR #923, opened 2023-01-04
extract all token details from es

CatalinFaurPaul:
## Reasoning
-  Instead of performing n vm queries for n tokens, perform scrollable requests on the ES, which should be much faster & reliable
  
## Proposed Changes
-  Retrieve from elastic all token properties

## How to test
-  /tokens
- /tokens/MEX-455c57 -> all token properties should be defined
- /collections/SUPERAPES-396a23 -> all collection properties should be defined


CatalinFaurPaul:
Tests passed

tanghel (approved):

PR #935, opened 2023-01-17
add owner attribute for providers

CatalinFaurPaul:
## Reasoning
-  on providers route /providers, owner attributes is missing
  
## Proposed Changes
-  add "owner" attribute

## How to test
- /providers -> should return all providers details + owner attribute defined
- /prividers/:address -> should return all provider details + owner attribute defined


CatalinFaurPaul:
Opened with another PR

PR #937, opened 2023-01-19
Allow herotags for smart contracts as well

tanghel:
## Reasoning
- In order to optimize performance, herotags were not fetched for smart contract, although the protocol allows it
  
## Proposed Changes
- since the number of smart contracts is relatively finite, we can allow fetching herotags for these as well

## How to test
- verify whether herotags are loaded for smart contracts

### DIFF
--- a/src/endpoints/accounts/account.service.ts
+++ b/src/endpoints/accounts/account.service.ts
@@ -109,9 +109,7 @@ export class AccountService {
         }
       }
 
-      if (!AddressUtils.isSmartContractAddress(address)) {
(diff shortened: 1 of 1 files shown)


gabrielmatei (approved):

CatalinFaurPaul (commented):
Approved

tanghel:
backtracked on this one, since:
- herotags are fetched from maiar id, which should be impossible for a smart contract to register itself on maiar id
- there is already the possibility to attach human-readable names to smart contracts with the help of assets

PR #947, opened 2023-01-26
Integrate new indexer mappings

tanghel:
## Reasoning
- Elasticsearch had an update where the mappings were improved, mainly `text` types were converted to `keyword`, which impacted the functioning of the API
  
## Proposed Changes
- create a config flag to specify whether to target the new and upcoming ES information, starting with the account collections query which did not need to use the `.keyword` specification

## How to test
- when targeting an updated indexer, such as `devnet-index.multiversx.com`, after setting the activation flag, account collections should start to be returned


gabrielmatei (approved):

CatalinFaurPaul:
Manual test passed

PR #959, opened 2023-02-07
Websocket event subscription

tanghel:
This PR will hold the integration between LiveEvents project and mx-api-service.
Initial steps are making sure the subscription model and the websocket work.

Throughout the series of commits, more features such as:

user authentication
redis adapter for multiple connections
user/transaction database
guards and validators
on/off toggle based on env variables
will be added, until the whole project is properly integrated.
Signed-off-by: Berciu Liviu [berciulivium@gmail.com](mailto:berciulivium@gmail.com)


src/common/auth/auth.service.ts
```diff
@@ -0,0 +1,158 @@
...
+
+    if (transaction) {
+      throw new HttpException(
+        `Transaction ${transactionAddress} already processed.`,
+        HttpStatus.BAD_REQUEST,
+      );
+    }
+
+    // Get transaction details
+    const txData = (
```
tanghel, line 67: here you can use TransactionService to fetch tx details

PR #970, opened 2023-02-15
Integrate native auth v1

tanghel:
## Reasoning
- Integrate latest native auth functionality

## Proposed Changes
- delete `JwtOrNativeAuthGuard` class and use the one in `sdk-nestjs`
- unified auth config under `features` with backwards compatibility
  
### Config Changes
old:
```yaml
api:
  auth: true
security:
  admins:
    - 'erd1...'
  jwtSecret: '12345'
```
new:
```yaml
  auth:
    enabled: true
    maxExpirySeconds: 86400
    acceptedOrigins:
      - 'https://xexchange.com'
    admins:
      - 'erd1...'
    jwtSecret: '12345'
```

## How to test
- test both legacy and new way of integrating auth in config which should work as before
- no more log messages if one or both auth checks fail
- in case one or both auth systems fail, check whether the response headers contain the reasons of error


gabrielmatei (dismissed):

gabrielmatei (approved):

CatalinFaurPaul:
Test passed with the following scenarios :green_circle: 

    - test both legacy and new way of integrating auth in config which should work as before 
    - no more log messages if one or both auth checks fail
    - in case one or both auth systems fail, check whether the response headers contain the reasons of error 
    - auth - enabled - false -> Throw forbidden resources :green_circle: 
    - auth - enabled - true -> Services are working

PR #990, opened 2023-03-06
Remove unnecessary logs

tanghel:
## Reasoning
- Error logs contained a lot of errors due to bad user input
  
## Proposed Changes
- Filter out errors for certain failures inside the proxy controller

## How to test
- `/vm-values/hex` with the body
```json
{
    "Args": [
        "fbe99a95a6925693679933db8eb7cb19ef4d187ac63db1b0204ad978e1263112"
    ],
    "FuncName": "getPublicKey",
    "ScAddress": "erd1qqqqqqqqqqqqqpgqjq0h2hy6ppj5yze2t3gr6kadvaahkljch77qynjfg8"
}
```
should not log any error
- `/address/:address/key/:key` should not return error for a key that doesn't exist

### DIFF
--- a/src/endpoints/proxy/proxy.controller.ts
+++ b/src/endpoints/proxy/proxy.controller.ts
@@ -49,7 +49,7 @@ export class ProxyController {
   @ApiExcludeEndpoint()
   async getAddressStorageKey(@Param('address', ParseAddressPipe) address: string, @Param('key') key: string) {
     // eslint-disable-next-line require-await
-    return await await this.gatewayGet(`address/${address}/key/${key}`, GatewayComponentRequest.addressStorage, undefined, async (error) => {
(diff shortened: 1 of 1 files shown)


src/endpoints/proxy/proxy.controller.ts
```diff
@@ -48,7 +48,14 @@ export class ProxyController {
   @Get('/address/:address/key/:key')
   @ApiExcludeEndpoint()
   async getAddressStorageKey(@Param('address', ParseAddressPipe) address: string, @Param('key') key: string) {
-    return await this.gatewayGet(`address/${address}/key/${key}`, GatewayComponentRequest.addressStorage);
+    // eslint-disable-next-line require-await
+    return await await this.gatewayGet(`address/${address}/key/${key}`, GatewayComponentRequest.addressStorage, undefined, async (error) => {
```
CatalinFaurPaul, line 52: remove 1 await from return

CatalinFaurPaul (approved):

PR #1002, opened 2023-03-10
Delegation shard

gfusee:
## Reasoning
- Delegation contract shard id can be computed instead of hardcoded in configs files
- Might not work if metachain id is not 4294967295, but it may be smarter to change the way shards are calculated inside mx-nestjs-sdk to take in account custom metachain id as optional parameter?
  
## Proposed Changes
- Compute the delegation contract shard id


tanghel (approved):

tanghel:
Thank you for the PR! It makes a lot of sense for the delegation shard id to be computed, one less config value to handle!

PR #1025, opened 2023-03-28
Latest block endpoint

dragos-rebegea:
## Reasoning
- Provide a dedicated endpoint for native auth client to query the latest block  
### Description:
This pull request adds a new endpoint to the API for retrieving details of the latest block on the blockchain. The endpoint is located at /blocks/latest and accepts optional query parameters to filter the block details based on the nonce frequency. If these query parameters are not provided, the endpoint returns information about the latest block that is at most one hour old.
### Endpoints Added:
- `GET /blocks/latest`

### Query Parameters:

- ttl: Compute the nonce frequency based on the TTL (time-to-live) value. If not specified, the latest block may be 1 hour old.
### Responses:
- 200 OK: Returns a BlockDetailed object containing the details of the latest block.
- 404 Not Found: If the requested block is not found.

### Steps:
1. Send a GET request to the /blocks/latest endpoint without any query parameters to retrieve details of the latest block that is at most one hour old.
2. Send a GET request to the /blocks/latest endpoint with the ttl query parameter set to a value (in seconds) to retrieve details of the latest block that is at most ttl seconds old.
### Example Usage:

1. GET /blocks/latest
2. GET /blocks/latest?ttl=3600

For each of these requests, verify that the API returns the expected response with the correct status code and block details. You can also test the behavior of the endpoint with different combinations of query parameters and values to ensure that it returns the expected results.


dragos-rebegea (commented):
:green_apple:

tanghel (approved):

PR #1043, opened 2023-04-12
Extra logging for GitHub validation

tanghel:
## Proposed Changes
- Provide more meaningful logs when peforming github.com keys validation in case of staking providers

## How to test
- Check whether logs appear when github.com keys not found, cannot be parsed or can be parsed successfully


CatalinFaurPaul:
:green_circle:

PR #1058, opened 2023-05-02
 add exchange filter 

cfaur09:
## Reasoning
- To improve accessibility and filtering of Mex pairs based on the exchange, we have added a new attribute, `exchange`, and introduced an `enum` to represent the possible exchange types.
- This implementation allows users to easily filter Mex pairs by their respective` exchange` and provides a more structured way to handle the data.
  
## Proposed Changes
- Added a new `exchange` attribute to the `MexPair` object
- Introduced the `MexPairExchangeType` enum to represent possible exchange types.
- Modified the `getPairInfo` method to assign the appropriate `exchange` value based on the pair type.
- Updated the `getMexPairs` and `getMexPairsCount` methods to accept an optional filter parameter of type `MexPairExchangeType`
- Implemented filtering logic in the `getMexPairs` and `getMexPairsCoun`t methods to return filtered results based on the provided filter value.
- Add new `ApiQuery` in `mex.controller.ts` to be able to apply the new exchange optional filter

## How to test
- `/mex/pairs?exchange=jungledex` -> should return ~21 mex pairs details with `exchange: "jungledex"` defined
- `/mex/pairs?exchange=xexchange` -> should return ~16 mex pairs details with `exchange: "xexchange"` defined
- `/mex/pairs/count?exchange=jungledex` -> should return ~21 mex pairs count
- `/mex/pairs/count?exchange=xexchange` -> should return ~16 mex pairs count

### DIFF
--- a/src/endpoints/mex/mex.pair.service.ts
+++ b/src/endpoints/mex/mex.pair.service.ts
@@ -9,6 +9,7 @@ import { MexPairType } from "./entities/mex.pair.type";
 import { MexSettingsService } from "./mex.settings.service";
 import { OriginLogger } from "@multiversx/sdk-nestjs";
 import { ApiConfigService } from "src/common/api-config/api.config.service";
+import { MexPairExchangeType } from "./entities/mex.pair.exchange.type";
 
 @Injectable()
 export class MexPairService {
@@ -27,8 +28,12 @@ export class MexPairService {
     await this.cachingService.setLocal(CacheInfo.MexPairs.key, pairs, Constants.oneSecond() * 30);
   }
 
-  async getMexPairs(from: number, size: number): Promise<any> {
-    const allMexPairs = await this.getAllMexPairs();
+  async getMexPairs(from: number, size: number, filter?: MexPairExchangeType): Promise<any> {
+    let allMexPairs = await this.getAllMexPairs();
+
+    if (filter) {
+      allMexPairs = allMexPairs.filter(pair => pair.exchange === filter);
+    }
 
     return allMexPairs.slice(from, from + size);
   }
@@ -51,8 +56,12 @@ export class MexPairService {
     );
   }
 
-  async getMexPairsCount(): Promise<number> {
+  async getMexPairsCount(filter?: MexPairExchangeType): Promise<number> {
     const mexPairs = await this.getAllMexPairs();
+
+    if (filter) {
+      return mexPairs.filter(pair => pair.exchange === filter).length;
+    }
 
     return mexPairs.length;
   }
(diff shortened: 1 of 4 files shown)


src/endpoints/mex/mex.pair.service.ts
```diff
@@ -128,10 +137,22 @@ export class MexPairService {
...
     const state = this.getPairState(pair.state);
     const type = this.getPairType(pair.type);
+
     if (!type || [MexPairType.unlisted].includes(type)) {
       return undefined;
     }
 
+    let exchange: MexPairExchange;
+
+    if (type === MexPairType.core || type === MexPairType.community ||
```
tanghel, line 147: you can use the includes function on an array to make this more compact

src/endpoints/mex/entities/mex.pair.exchange.ts
```diff
@@ -0,0 +1,23 @@
+import { registerEnumType } from "@nestjs/graphql";
+
+export enum MexPairExchange {
+  xexchange = 'xexchange',
+  jungledex = 'jungledex',
+  none = 'none'
```
tanghel, line 6: unknown instead of none should be a better name

src/endpoints/mex/mex.pair.service.ts
```diff
@@ -51,9 +56,13 @@ export class MexPairService {
     );
   }
 
-  async getMexPairsCount(): Promise<number> {
+  async getMexPairsCount(filter?: MexPairExchange): Promise<number> {
```
tanghel, line 59: please create a MexPairsFilter class to hold the exchange filter, so that it's in line with other functions and also can be easily extended

src/endpoints/mex/entities/mex.pair.exchange.ts
cfaur09: done

src/endpoints/mex/mex.pair.service.ts
cfaur09: done

src/endpoints/mex/mex.pair.service.ts
```diff
@@ -56,11 +58,11 @@ export class MexPairService {
     );
   }
 
-  async getMexPairsCount(filter?: MexPairExchange): Promise<number> {
+  async getMexPairsCount(filter?: MexPairsFilter): Promise<number> {
     const mexPairs = await this.getAllMexPairs();
 
-    if (filter) {
-      return mexPairs.filter(pair => pair.exchange === filter).length;
+    if (filter && filter.exchange) {
```
tanghel, line 64: should be a common function that applies the filter, to re-use the same logic when adding multiple filters

PR #1061, opened 2023-05-05
add inTransit transaction field

cfaur09:
## Reasoning
- The current implementation of the getTransaction method does not provide information about whether a transaction is in transit or not.
- In transit transactions are those with a miniBlockHash and a status of "pending".
 - Including this information can be helpful for users to better understand the state of their transactions.
  
## Proposed Changes
- Add a new field inTransit to the TransactionDetailed class as a boolean.
- Modify the tryGetTransactionFromGateway method to set the inTransit field to true if miniBlockHash is not undefined and status is "pending".

## How to test ( devnet )
- Make a new transaction ( ESDT Transfer ) and within 5 seconds make a request to /transactions/:txHash and see if `inTransit` fields is defined as true, but check if miniblockHash is defined and status: pending

### DIFF
--- a/src/endpoints/transactions/transaction.get.service.ts
+++ b/src/endpoints/transactions/transaction.get.service.ts
@@ -194,6 +194,7 @@ export class TransactionGetService {
         logs: transaction.logs,
         guardianAddress: transaction.guardian,
         guardianSignature: transaction.guardianSignature,
+        inTransit: transaction.miniblockHash !== undefined && transaction.status === 'pending',
       };
 
       return ApiUtils.mergeObjects(new TransactionDetailed(), result);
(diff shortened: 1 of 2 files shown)


src/endpoints/transactions/transaction.get.service.ts
```diff
@@ -194,6 +194,7 @@ export class TransactionGetService {
         logs: transaction.logs,
         guardianAddress: transaction.guardian,
         guardianSignature: transaction.guardianSignature,
+        inTransit: transaction.miniblockHash !== undefined && transaction.status === 'pending',
```
tanghel, line 197: you should use the enum value if possible

PR #1078, opened 2023-06-12
add remainingUnbondPeriod

cfaur09:
## Reasoning
- Not all nodes should have the `remainingUnBondPeriod` attribute, only those with the status `NodeStatus.leaving`. For other nodes, this attribute should be undefined, not 0.
  
## Proposed Changes
- Added a function `applyNodeUnbondingPeriods` in the NodeService. This function assigns each node's `remainingUnBondPeriod` attribute by calling `keysService.getKeyUnbondPeriod` if the node status is NodeStatus.leaving. If not, `remainingUnBondPeriod` is set to undefined.
- Updated the `getAllNodesRaw()` function in `NodeService` to call the new `applyNodeUnbondingPeriods` function, adding remainingUnBondPeriod to all nodes with the status NodeStatus.leaving in the process.

## How to test ( DEVNET)
- Add node -> Stake node -> Unstake node -> all from web wallet
- After this stepts, take the bls key and add it to nodes/:bls
- Response body should contains `remainingUnbondPeriod`

example:
- `nodes/204ee6c9a68a6a0d5a4af5426d5a34b1ff0a5c62d0f93ee09aabbc54ad7f864b1f7c69d6d832d98425ce8626884cb611248e1004c02bb95acb821231195f388c8cc2dd7cb79c4f35d77bfb814dc5e5e297013252622320374eb744beb26a4794` -> should contains `remainingUnbondPeriod`

### DIFF
--- a/src/endpoints/nodes/node.service.ts
+++ b/src/endpoints/nodes/node.service.ts
@@ -272,14 +272,18 @@ export class NodeService {
   }
 
   private async applyNodeUnbondingPeriods(nodes: Node[]): Promise<void> {
-    await Promise.all(nodes.map(async node => {
-      if (node.status === NodeStatus.leaving) {
-        const keyUnbondPeriod = await this.keysService.getKeyUnbondPeriod(node.bls);
-        node.remainingUnBondPeriod = keyUnbondPeriod?.remainingUnBondPeriod;
-      } else {
+    const leavingNodes = nodes.filter(node => node.status === NodeStatus.leaving);
+
+    await Promise.all(leavingNodes.map(async node => {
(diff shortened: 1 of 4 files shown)


src/endpoints/nodes/node.service.ts
```diff
@@ -269,6 +271,17 @@ export class NodeService {
     }
   }
 
+  private async applyNodeUnbondingPeriods(nodes: Node[]): Promise<void> {
+    await Promise.all(nodes.map(async node => {
```
tanghel, line 275: Promise.all should be made only on the leaving nodes, because only then an await will be performed

cfaur09: fixed

tanghel (approved):

PR #1110, opened 2023-08-29
add withLatestInfo parameter

cfaur09:
## Reasoning
- In the current implementation, for the provider service, the information is cached, not being able to see the values ​​of certain fields in real time 
  
## Proposed Changes
1. Adding a flag called `withLatestInfo` which, when `true`, calls will be made to 2 `vm-queries` that will update certain fields
`serviceFee`, `automaticActivation`, `checkCapOnRedelegate`, `delegationCap`, `locked`
- vm-queries: `getContractConfig` and `getTotalActiveStake`
- Adding `withLatestInfo` as an API query parameter allows consumers of the API to specify whether they want to include the latest information about providers or not.
- Adding an options object to hold `withIdentityInfo` and `withLatesttInfo` flags.

## How to test ( TESTNET )
- `providers?owner=erd1czgky3rcqggezg6zan5k2c9sfnwlh644jc26t6vud256ack9xaaqvhqddm&withLatestInfo=true` -> try to stake to contract `erd1qqqqqqqqqqqqqqqpqqqqqqqqqqqqqqqqqqqqqqqqqqqqqx8llllsxavffq` where owner is `erd1czgky3rcqggezg6zan5k2c9sfnwlh644jc26t6vud256ack9xaaqvhqddm` and after transaction is success, you can observe that the `locked` value is updated with the new value. To compare, make the same request to `testnet-api.multiversx.com/providers?owner=erd1czgky3rcqggezg6zan5k2c9sfnwlh644jc26t6vud256ack9xaaqvhqddm&withLatestInfo=true`

- To be able to test all the fields, you need to create your own delegation contract and change all the values to observe the change s in your local API

### DIFF
--- a/src/endpoints/providers/provider.service.ts
+++ b/src/endpoints/providers/provider.service.ts
@@ -146,10 +146,6 @@ export class ProviderService {
       return bSort - aSort;
     });
 
-
-
-    // providers = providers.filter(provider => provider.numNodes > 0);
-
     for (const provider of providers) {
       if (!provider.identity) {
         continue;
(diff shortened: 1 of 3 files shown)


src/endpoints/providers/provider.service.ts
```diff
@@ -448,4 +473,17 @@ export class ProviderService {
...
+
+  private async getTotalActiveStake(address: string): Promise<string> {
+    const [activeStake] = await this.vmQueryService.vmQuery(
+      address,
+      'getTotalActiveStake',
+    );
+    return this.numberDecode(activeStake);
+  }
+
+  numberDecode(encoded: string): string {
```
tanghel, line 485: instead of creating this function, please use BinaryUtils

tanghel (approved):

PR #1112, opened 2023-08-29
add complexity to withLatestInfo filter

cfaur09:
## Proposed Changes
- 
- 
- 

## How to test
- `/providers?withLatestInfo=true` -> should throw `BadRequestException` because `owner` filter is not defined
- `providers?owner=erd1fx5t2nwq4fh9jws5xqfl85hr0l8tuqks9sr7ut9wrpkp7dugzxnqyksfyg&withLatestInfo=true` -> should return provider info based on `owner` filter and `withLatestInfo` filter applied

### DIFF
--- a/src/endpoints/providers/entities/provider.query.options.ts
+++ b/src/endpoints/providers/entities/provider.query.options.ts
@@ -9,7 +9,7 @@ export class ProviderQueryOptions {
   withIdentityInfo?: boolean = false;
 
   static applyDefaultOptions(owner: string | undefined, options: ProviderQueryOptions): ProviderQueryOptions {
-    if (options.withLatestInfo === true && owner === undefined && !owner) {
+    if (options.withLatestInfo === true && !owner) {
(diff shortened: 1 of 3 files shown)


src/endpoints/providers/entities/provider.query.options.ts
```diff
@@ -0,0 +1,18 @@
...
+export class ProviderQueryOptions {
+  constructor(init?: Partial<ProviderQueryOptions>) {
+    Object.assign(this, init);
+  }
+
+  withLatestInfo?: boolean = false;
+  withIdentityInfo?: boolean = false;
+
+  static applyDefaultOptions(owner: string | undefined, options: ProviderQueryOptions): ProviderQueryOptions {
+    if (options.withLatestInfo === true && owner === undefined) {
```
tanghel, line 12: `&& !owner` to also take into account situations where it is string empty and it may bypass this check

tanghel (approved):

PR #1126, opened 2023-09-21
add isSmartContract filter to be able to return a list of smart contr…

cfaur09:
## Proposed Changes
- Add `isSmartContract` filter to be able to return a list of smart contracts
- Add new query in elastic based on `currentOwner`
- Add in`accounts.controller.ts` new `ApiQuery` / `Query` -> `isSmartContract`

## How to test
- `/accounts?isSmartContract=true` -> returned list should contain only smart contracts
- `/accounts?isSmartContract=false` -> returned list should not contain smart contracts
- `/accounts` -> returned list should contain smart contracts / user accounts

### DIFF
--- a/src/endpoints/accounts/account.service.ts
+++ b/src/endpoints/accounts/account.service.ts
@@ -336,18 +336,6 @@ export class AccountService {
     for (const account of accounts) {
       account.shard = AddressUtils.computeShard(AddressUtils.bech32Decode(account.address), shardCount);
       account.assets = assets[account.address];
-
-      if (filter.isSmartContract) {
-        const deployedAt = await this.getAccountDeployedAt(account.address);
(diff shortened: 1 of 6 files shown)


src/endpoints/accounts/account.service.ts
```diff
@@ -293,7 +293,7 @@ export class AccountService {
   }
 
   async getAccounts(queryPagination: QueryPagination, filter: AccountFilter): Promise<Account[]> {
-    if (!filter.ownerAddress && !filter.sort && !filter.order) {
+    if (!filter.ownerAddress && !filter.sort && !filter.order && !filter.isSmartContract) {
```
tanghel, line 296: `filter.isSmartContract === undefined`

tanghel (approved):

PR #1147, opened 2023-11-09
Added a few tests to extend the current test suite for ApiChecker.

ramona-lobont:
## Reasoning
- 
- 
- 
  
## Proposed Changes
- 
- 
- 

## How to test
- 
- 
-


src/plugins.template/plugin.module.ts.template
cfaur09, line 1: this should not be deleted
Please revert

PR #1157, opened 2023-12-04
added e2e tests for accounts controller

ramonalobont:
## Reasoning
- 
- 
- 
  
## Proposed Changes
- 
- 
- 

## How to test
- 
- 
-

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "elrond-nestjs-api",
   "version": "0.0.1",
-  "lockfileVersion": 3,
+  "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
@@ -114,6 +114,15 @@
         "webpack": "^5.73.0"
       }
     },
+    "node_modules/@aashutoshrathi/word-wrap": {
+      "version": "1.2.6",
+      "resolved": "https://registry.npmjs.org/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz",
+      "integrity": "sha512-1Yjs2SvM8TflER/OD3cOjhWWOZb58A2t7wpE2S9XfBYTiIl+XFhQG2bjy4Pu1I+EAlCNUzRDYDdFwFYUKvXcIA==",
+      "dev": true,
+      "engines": {
+        "node": ">=0.10.0"
+      }
+    },
     "node_modules/@acuminous/bitsyntax": {
       "version": "0.1.2",
       "resolved": "https://registry.npmjs.org/@acuminous/bitsyntax/-/bitsyntax-0.1.2.tgz",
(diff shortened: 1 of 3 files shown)


src/plugins.template/plugin.module.ts.template
cfaur09, line 1: Please undo plugins. No need to be deleted

package-lock.json
cfaur09, line 1: If you didn't update any dependencies, please undo

src/test/integration/controllers/accounts.controller.e2e-spec.ts
```diff
@@ -0,0 +1,252 @@
...
+                    expect(res.body.message).toEqual("Validation failed for argument 'ownerAddress' (a bech32 address is expected)");
+                });
+        });
+
+        [
+            {
+                filter: 'sort',
+                value: 'balance',
+            },
+        ].forEach(({ filter, value }) => {
```
cfaur09, line 98: This test can contain also the below test, no need to create another test just to test another sort type

src/test/integration/controllers/accounts.controller.e2e-spec.ts
```diff
@@ -0,0 +1,252 @@
...
+                        .expect(200)
+                        .then(res => {
+                            expect(res.body).toBeDefined();
+                            expect(res.body[0].timestamp).toBeGreaterThanOrEqual(res.body[1].timestamp);
+                        });
+                });
+            });
+        });
+
+        test.each`
```
cfaur09, line 135: This test is good, but I consider that you can remove the following test cases because are the same:

- should return 25 accounts available on blockchain
- should return 1 account details
- should return 7 accounts

src/test/integration/controllers/accounts.controller.e2e-spec.ts
```diff
@@ -0,0 +1,252 @@
...
+                        .expect(200)
+                        .then(res => {
+                            expect(res.body).toHaveLength(size);
+                        });
+                }
+            );
+    });
+
+    describe('/accounts/count', () => {
+        it('should return count of all blocks from all shards', async () => {
```
cfaur09, line 159: This is not the correct test case description. You need to add description for accounts

src/test/integration/controllers/accounts.controller.e2e-spec.ts
```diff
@@ -0,0 +1,252 @@
...
+                        .then(res => {
+                            expect(+res.text).toBeGreaterThanOrEqual(count);
+                        });
+                });
+            });
+        });
+    });
+
+    describe('/accounts/{address}', () => {
+        it('should return account details for a given address', async () => {
```
cfaur09, line 204: Please retake this test case because, sometimes will throw fail
Especially for the res.body.balance, because this balance can change over time

src/test/integration/controllers/accounts.controller.e2e-spec.ts
```diff
@@ -0,0 +1,252 @@
...
+                    expect(res.body.isGuarded).toStrictEqual(false);
+                    expect(res.body.isPayable).toStrictEqual(false);
+                    expect(res.body.deployTxHash).toStrictEqual('215e24332cc2734e499be534268a7badd13dcf82735cfc0ab571fe09d9fe0804');
+                    expect(res.body.deployedAt).toStrictEqual(1689340068);
+                });
+        });
+    });
+
+    describe('Validations', () => {
+        it('should return 400 Bad Request for an invalid address', async () => {
```
cfaur09, line 236: This test is the same with the following test:
- should return 400 Bad Request if owner address is not correct

PR #1184, opened 2024-01-17
added sender and receiver usernames

bogdan-rosianu:
## Reasoning
- senderUsername and receiverUsername fields were missing from the transactions endpoints
  
## Proposed Changes
- add missing fields

## How to test
- on testnet: http://localhost:3001/transactions/050eaebefb288e698dcc8ba129781ea0d1c491a1519c1e4f9ba937b3444fb646 should return the receiverUsername as well


src/endpoints/transactions/transaction.get.service.ts
```diff
@@ -116,6 +116,14 @@ export class TransactionGetService {
         }
       }
 
+      if (transaction.senderUserName) {
+        transactionDetailed.senderUsername = UsernameUtils.extractUsernameFromRawBase64(transaction.senderUserName);
```
tanghel, line 120: let's prepare for the upcoming ES upgrade and add support for both `senderUserName` and `senderUsername` fields directly

bogdan-rosianu: done

tanghel (approved):

PR #1207, opened 2024-02-22
status checker initialize thresholds

cfaur09:
## Reasoning
- Enhance adaptability across different environments (`mainnet`, `devnet`, `testnet`).
- Streamline configuration by embedding threshold settings within the service logic.
  
## Proposed Changes
- Adding a new method, `initializeThresholds`, in the `StatusCheckerService`, dynamically setting thresholds based on the network environment identified through apiConfigService. This method adjusts thresholds for:
- `Token, nodes, and providers counts.`
- `Token supply, assets, accounts, and transactions.`
- `Node validators count.`

## How to test
- Enable statusChecker flag and check for every network that all the values are set correctly


src/crons/status.checker/status.checker.service.ts
```diff
@@ -40,147 +49,193 @@ export class StatusCheckerService {
...
     private readonly economicService: NetworkService,
-  ) { }
+    private readonly apiConfigService: ApiConfigService
+  ) {
+    this.initializeThresholds();
+  }
+
+  private initializeThresholds() {
+    switch (this.apiConfigService.getNetwork()) {
+      case 'mainnet':
```
tanghel, line 59: please move these settings in the config files, as an object with attributes 🙏

cfaur09: done

tanghel (approved):

tanghel (approved):

PR #1242, opened 2024-04-18
add lowLiquidityThresholdPercent property

cfaur09:
## Reasoning
- Need to ensure fixed display value for liquidity conditions
- Simplify the codebase for easier maintenance and readability
- Improve predictability in UI behavior regarding liquidity thresholds
  
## Proposed Changes
- Set `lowLiquidityThresholdPercent` to a fixed value of `0.5%` for tokens classified as having low liquidity.

## How to test
- `tokens/SHARD-99a172` -> `lowLiquidityThresholdPercent` should be defined with value 0.5
- `tokens` -> search for `lowLiquidityThresholdPercent` and check if specific tokens contains `lowLiquidityThresholdPercent`

### DIFF
--- a/src/endpoints/tokens/entities/token.ts
+++ b/src/endpoints/tokens/entities/token.ts
@@ -155,11 +155,11 @@ export class Token {
   @ApiProperty({ type: Number, nullable: true })
   totalVolume24h: number | undefined = undefined;
 
-  @Field(() => Boolean, { description: 'If the liquidity to market cap ratio is less than 1%, we consider it as low liquidity.', nullable: true })
+  @Field(() => Boolean, { description: 'If the liquidity to market cap ratio is less than 0.5%, we consider it as low liquidity.', nullable: true })
   @ApiProperty({ type: Boolean, nullable: true })
   isLowLiquidity: boolean | undefined = undefined;
 
-  @Field(() => Number, { description: 'If the liquidity to market cap ratio is less than 1%, we consider it as low liquidity and display threshold percent .', nullable: true })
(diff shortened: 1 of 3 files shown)


src/endpoints/tokens/entities/token.ts
```diff
@@ -155,7 +155,11 @@ export class Token {
   @ApiProperty({ type: Number, nullable: true })
   totalVolume24h: number | undefined = undefined;
 
-  @Field(() => Boolean, { description: 'If the liquidity to market cap ratio is less than 1%, we consider it as low liquidity.', nullable: true })
+  @Field(() => Boolean, { description: 'If the liquidity to market cap ratio is less than 0.5%, we consider it as low liquidity.', nullable: true })
   @ApiProperty({ type: Boolean, nullable: true })
   isLowLiquidity: boolean | undefined = undefined;
+
+  @Field(() => Number, { description: 'If the liquidity to market cap ratio is less than 0.5%, we consider it as low liquidity and display threshold percent .', nullable: true })
```
bogdan-rosianu, line 162: might remove references to '0.5%' since we can change it

cfaur09: I think it would be good to leave the value in the description and if we change it in the future, we will change it in the ApiProperty as well

bogdan-rosianu: ok 👍

bogdan-rosianu (approved):

tanghel (approved):

PR #1249, opened 2024-05-14
add default icons for exchange contracts

cfaur09:
## Reasoning
- We need to add these fields for every exchange contract.
- These fields will be used by the frontend (explorer, wallet) to display appropriate icons for the exchange contracts.

## Proposed Changes
- To avoid repetitive code when setting iconSvg and iconPng for AccountAssets created a helper function to set AccountAssets fields
- Add for every exchange contract `iconSvg` and `iconPng`

## How to test
- `accounts/erd1qqqqqqqqqqqqqpgqvc7gdl0p4s97guh498wgz75k8sav6sjfjlwqh679jy` -> check if assets contains also icons


cfaur09:
use this PR instead https://github.com/multiversx/mx-api-service/pull/1250

PR #1251, opened 2024-05-15
add the nft processing queue name in configs

cfaur09:
## Reasoning
- At the moment the queue: 'api-process-nfts' is hardcoded in the code and the same rabbitmq cannot be used by 2 different api setups
  
## Proposed Changes
- Add the queue name for nft process queue in configs


danielailie (dismissed):

bogdan-rosianu (dismissed):

IosifGabriel:
Hello, I still think there needs to be done some concatetation for the deadLetterExchange as it remained hardcoded

https://github.com/multiversx/mx-api-service/blob/5b0d9f1a7a976cb4c4e44116da8c1572d6337ac6/src/main.ts#L129
https://github.com/multiversx/mx-api-service/blob/5b0d9f1a7a976cb4c4e44116da8c1572d6337ac6/src/queue.worker/nft.worker/nft.worker.module.ts#L35

cfaur09:
> Hello, I still think there needs to be done some concatetation for the deadLetterExchange as it remained hardcoded
> 
> https://github.com/multiversx/mx-api-service/blob/5b0d9f1a7a976cb4c4e44116da8c1572d6337ac6/src/main.ts#L129
> 
> 
> https://github.com/multiversx/mx-api-service/blob/5b0d9f1a7a976cb4c4e44116da8c1572d6337ac6/src/queue.worker/nft.worker/nft.worker.module.ts#L35

Solved 🟢

bogdan-rosianu (approved):

tanghel (commented):
I would make here a feature called "processNfts" and add there both the normal queue and the dead letter queue name. when reading the activation flag, try to read it first from features and then also from the "old" location to maintain backwards compatibility. update all config files with the new location. in this way we have all process nft-related functionality grouped together

bogdan-rosianu (dismissed):

bogdan-rosianu (approved):

tanghel (approved):

PR #1265, opened 2024-05-22
add node waiting epochs left proxy

cfaur09:
## Reasoning
- new node endpoint is available in gateway.multiversx.com
  
## Proposed Changes
- Proxy forward node/waiting-epochs-left/:bls endpoint

## How to test(mainnet)
- `node/waiting-epochs-left/ccea9c6d5bb3833dcb90358d9585d1dddb113523d049ec93762774e620c38dabbf1868fb7b61712c3e573d788203270e0efffc97f33af38e7f46b216d04694b13c49e91f33c1bfda000634fde3557c65390065f0d5a8a211020dfa36f5c72f90` -> should return ` "epochsLeft": 3`

### DIFF
--- a/src/endpoints/proxy/gateway.proxy.controller.ts
+++ b/src/endpoints/proxy/gateway.proxy.controller.ts
@@ -6,7 +6,7 @@ import { GatewayService } from "src/common/gateway/gateway.service";
 import { Response, Request } from "express";
 import { GatewayComponentRequest } from "src/common/gateway/entities/gateway.component.request";
 import { PluginService } from "src/common/plugins/plugin.service";
-import { Constants, ParseAddressPipe, ParseBlockHashPipe, ParseIntPipe, ParseTransactionHashPipe } from "@multiversx/sdk-nestjs-common";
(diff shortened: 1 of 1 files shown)


gabrielmatei (approved):

src/endpoints/proxy/gateway.proxy.controller.ts
```diff
@@ -248,6 +248,13 @@ export class GatewayProxyController {
     }
   }
 
+  @Get('/node/waiting-epochs-left/:bls')
+  async getNodeWaitingEpochsLeft(
+    @Param('bls', ParseBlsHashPipe) bls: string,
```
bogdan-rosianu, line 253: offtopic: the BLS key is not a hash, but rather a hex encoding of a public key of a validator key.

bogdan-rosianu (approved):

src/endpoints/proxy/gateway.proxy.controller.ts
tanghel: yep, we need to do a breaking change in sdk-nestjs for that. let's put it on the task list and address this later on

tanghel (approved):

PR #1300, opened 2024-08-06
Support for new nft types

tanghel:
## Reasoning
- In the upcoming protocol release, we have additional NFT collection types related to dynamic NFTs and NFT v2, which need accomodation at api level
  
## Proposed Changes
- In order to make the transition as smooth as possible and avoid breaking all dapps that perform queries on the collections and nfts endpoints, the new types are mapped on the existing types `NonFungibleESDT`, `SemiFungibleESDT` and `MetaESDT`, with a `subType` attribute differentiating the new types

## How to test (testnet)
- `/collections` should return also collections of subType `NonFungibleESDTv2`
- `/collections?type=NonFungibleESDT` should return also collections of subType `NonFungibleESDTv2`
- `/nfts` should return also NFTs of subType `NonFungibleESDTv2`
- `/nfts?type=NonFungibleESDT` should return also NFTs of subType `NonFungibleESDTv2`
- `/nfts/TEST-41f2aa-01` should have type `NonFungibleESDT` and subType `NonFungibleESDTv2`
- `/accounts/erd1e7qgngs0sq2f9fkyqa0ddvy52a66t6yrv9jd09afrn9cqm0lc4cquhm7ug/nfts` should return also NFTs of subType `NonFungibleESDTv2`
- `/accounts/erd1e7qgngs0sq2f9fkyqa0ddvy52a66t6yrv9jd09afrn9cqm0lc4cquhm7ug/nfts/TEST-41f2aa-01` should return the NFT of type `NonFungibleESDTv2`


gabrielmatei (dismissed):

src/utils/cache.info.ts
```diff
@@ -194,7 +194,7 @@ export class CacheInfo {
 
   static EsdtProperties(identifier: string): CacheInfo {
     return {
-      key: `esdt:${identifier}`,
+      key: `esdt:v2:${identifier}`,
```
bogdan-rosianu, line 197: I think this is in order to invalidate old cache keys. but we should revert it later

src/endpoints/esdt/esdt.service.ts
```diff
@@ -235,6 +252,22 @@ export class EsdtService {
       delete tokenProps.NFTCreateStopped;
     }
 
+    switch (elasticProperties.type) {
```
bogdan-rosianu, line 255: duplicated code with L173-L187. Might extract a function in a helper package or something

src/common/indexer/elastic/elastic.indexer.helper.ts
```diff
@@ -76,7 +76,15 @@ export class ElasticIndexerHelper {
   public buildCollectionRolesFilter(filter: CollectionFilter, address?: string): ElasticQuery {
     let elasticQuery = ElasticQuery.create();
     elasticQuery = elasticQuery.withMustNotExistCondition('identifier')
-      .withMustMultiShouldCondition([NftType.MetaESDT, NftType.NonFungibleESDT, NftType.SemiFungibleESDT], type => QueryType.Match('type', type));
+      .withMustMultiShouldCondition([
+        NftType.MetaESDT,
```
bogdan-rosianu, line 80: you could have used `Object.values(NftType)` instead of stating each member of the enum. also future proof

src/common/indexer/elastic/elastic.indexer.helper.ts
```diff
@@ -127,14 +135,38 @@ export class ElasticIndexerHelper {
     }
 
     if (filter.excludeMetaESDT === true) {
-      elasticQuery = elasticQuery.withMustMultiShouldCondition([NftType.NonFungibleESDT, NftType.SemiFungibleESDT], type => QueryType.Match('type', type));
+      elasticQuery = elasticQuery.withMustMultiShouldCondition([
+        NftType.NonFungibleESDT,
```
bogdan-rosianu, line 139: we could have defined some subgroups, something like:
```
export const semiFungibleGroup = [
  NftType.SemiFungibleESDT,
  NftType.DynamicSemiFungibleESDT,
];
export const nonFungibleGroup = [
  NftType.NonFungibleESDT,
  NftType.NonFungibleESDTv2,
  NftType.DynamicNonFungibleESDT,
];
```

This part of the code can become:
```
semiFungibleGroup.concat(dynamicGroup)
```
which is also future proof

src/endpoints/collections/collection.service.ts
```diff
@@ -80,7 +82,26 @@ export class CollectionService {
   }
 
   applyPropertiesToCollectionFromElasticSearch(nftCollection: NftCollection, indexedCollection: Collection) {
-    nftCollection.type = indexedCollection.type as NftType;
+    switch (indexedCollection.type) {
```
bogdan-rosianu, line 85: also here, we'll need to update these conditions for each new type.

src/utils/cache.info.ts
tanghel: This should normally as it is since we have made some incompatible changes and the ttl is one week

src/endpoints/esdt/esdt.service.ts
tanghel: done, thanks !

src/common/indexer/elastic/elastic.indexer.helper.ts
tanghel: very good idea, implemented it for now as a member of the class itself

bogdan-rosianu (approved):

cfaur09 (approved):

PR #1306, opened 2024-08-14
Nodes and tokens fetch from external api

tanghel:
## Reasoning
- In order to avoid hitting rate limits on ES / gateway, we can specify a set of features to activate fetching of nodes and tokens from an external API, so that we don't perform a lot of requests to the downstream resources and hitting 429 on them
  
## Proposed Changes
- Add external API fetch for the nodes and token routes
- Add 2 features in the config file for nodesFetch and tokenFetch with the fields enabled and serviceURL

## How to test
- when tokenFetch is enabled, calling getAllTokens() should make 2 GET calls through apiService at the serviceURL parameter and not call any other function
- when nodesFetch is enabled, calling getAllNodes() should make 2 GET calls through apiService at the serviceURL parameter and not call any other function

### DIFF
--- a/config/config.mainnet.yaml
+++ b/config/config.mainnet.yaml
@@ -97,7 +97,13 @@ features:
     deadLetterQueueName: 'api-process-nfts-dlq'
   tps:
     enabled: true
-    maxLookBehindNonces: 100  
+    maxLookBehindNonces: 100
+  nodesFetch:
+    enabled: true
+    serviceUrl: 'https://mainnet-api.multiversx.com'
+  tokensFetch:
+    enabled: true
+    serviceUrl: 'https://mainnet-api.multiversx.com'
 image:
   width: 600
   height: 600
--- a/config/config.mainnet.yaml
+++ b/config/config.mainnet.yaml
@@ -100,10 +100,10 @@ features:
     maxLookBehindNonces: 100
   nodesFetch:
     enabled: true
-    serviceUrl: 'https://mainnet-api.multiversx.com'
+    serviceUrl: 'https://api.multiversx.com'
   tokensFetch:
     enabled: true
-    serviceUrl: 'https://mainnet-api.multiversx.com'
+    serviceUrl: 'https://api.multiversx.com'
 image:
   width: 600
   height: 600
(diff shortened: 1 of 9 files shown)


config/config.mainnet.yaml
```diff
@@ -97,7 +97,13 @@ features:
     deadLetterQueueName: 'api-process-nfts-dlq'
   tps:
     enabled: true
-    maxLookBehindNonces: 100  
+    maxLookBehindNonces: 100
+  nodesFetch:
+    enabled: true
+    serviceUrl: 'https://mainnet-api.multiversx.com'
```
gabrielmatei, line 103: replace with https://api.multiversx.com

config/config.mainnet.yaml
```diff
@@ -97,7 +97,13 @@ features:
...
+  tokensFetch:
+    enabled: true
+    serviceUrl: 'https://mainnet-api.multiversx.com'
```
gabrielmatei, line 106: replace with https://api.multiversx.com

src/endpoints/nodes/node.service.ts
```diff
@@ -392,6 +398,27 @@ export class NodeService {
     return nodes;
   }
 
+  private async getAllNodesFromApi(): Promise<Node[]> {
+    try {
+      const requestNodesCount = await this.apiService.get(this.apiConfigService.getNodesFetchServiceUrl() + '/nodes/count');
```
gabrielmatei, line 403: ``${this.apiConfigService.getNodesFetchServiceUrl()}/nodes/count``

src/endpoints/nodes/node.service.ts
```diff
@@ -392,6 +398,27 @@ export class NodeService {
...
+      };
+
+      const requestNodes = await this.apiService.get(this.apiConfigService.getNodesFetchServiceUrl() + '/nodes', requestUrlParams);
+      const nodes = requestNodes.data;
+
+      return nodes;
+    } catch (error) {
+      this.logger.error('An unhandled error occurred when getting nodes from API');
+      this.logger.error(error);
+      return [];
```
gabrielmatei, line 418: throw error

src/endpoints/tokens/token.service.ts
```diff
@@ -721,6 +722,10 @@ export class TokenService {
   async getAllTokensRaw(): Promise<TokenDetailed[]> {
     this.logger.log(`Starting to fetch all tokens`);
```
gabrielmatei, line 723: move this log after if

cfaur09 (approved):

PR #1309, opened 2024-08-16
Integration of K6 into API Github workflow

MogageNicolae:
## Reasoning
- 
  
## Proposed Changes
- 

## How to test
-


.github/workflows/load-tests.yml
```diff
@@ -0,0 +1,182 @@
+name: Load Tests
+
+on: [pull_request]
```
cfaur09, line 3: we need to be sure that this action is running only on main and development branch ( on push )

cfaur09 (approved):

PR #1331, opened 2024-09-12
tokens fetch from external api unit tests

GuticaStefan:
## Reasoning
- no testcase implemented for the case when `isTokensFetchFeatureEnabled` is false
  
## Proposed Changes
- create new unit test for the `getAllTokens` when `isTokensFetchFeatureEnabled` is false
- adjust also the unit test for the `getAllTokens` when `isTokensFetchFeatureEnabled` is true

## How to test
- `npm run test tokens.spec.ts`

### DIFF
--- a/src/test/unit/services/tokens.spec.ts
+++ b/src/test/unit/services/tokens.spec.ts
@@ -30,6 +30,7 @@ import { Token } from "src/endpoints/tokens/entities/token";
 import { NftCollection } from "src/endpoints/collections/entities/nft.collection";
 import { EsdtSupply } from "src/endpoints/esdt/entities/esdt.supply";
 import { TokenDetailed } from "src/endpoints/tokens/entities/token.detailed";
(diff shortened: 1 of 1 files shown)


src/test/unit/services/tokens.spec.ts
```diff
@@ -617,18 +631,118 @@ describe('Token Service', () => {
...
+            canFreeze: collection.canFreeze,
+            canPause: collection.canPause,
+            canTransferNftCreateRole: collection.canTransferNftCreateRole,
+            canWipe: collection.canWipe,
+            canAddSpecialRoles: collection.canAddSpecialRoles,
+            canChangeOwner: collection.canChangeOwner,
+            canUpgrade: collection.canUpgrade,
+          }));
+        });
+        console.log(mockTokens);
```
cfaur09, line 713: remove console.log

cfaur09 (approved):

PR #1334, opened 2024-09-19
add mex token charts

cfaur09:
## Reasoning
- Token price history data are available on xexchange graphql
- Hourly and daily resolutions offer different granularities for analysis
  
## Proposed Changes
- Add new endpoints for retrieving token price charts
- Implement hourly resolution for the last 24 hours of price data
- Implement daily resolution for historical price data with a specified start date

## How to test
- `<api>/mex/tokens/prices/daily/HTM-f51d55?after=1724054986`
- `<api>/mex/tokens/prices/hourly/HTM-f51d55`


src/endpoints/mex/mex.token.charts.service.ts
```diff
@@ -0,0 +1,106 @@
...
+      async () => await this.getTokenPricesHourResolutionRaw(tokenIdentifier),
+      CacheInfo.TokenHourChar(tokenIdentifier).ttl,
+    );
+  }
+
+  async getTokenPricesHourResolutionRaw(tokenIdentifier: string): Promise<MexTokenChart[]> {
+    const tokenExists = await this.checkTokenExists(tokenIdentifier);
+
+    if (!tokenExists) {
+      this.logger.error(`Token ${tokenIdentifier} does not exist or has no trading pair`);
```
bogdan-rosianu, line 32: does it worth logging this for each user bad request?

src/utils/cache.info.ts
```diff
@@ -155,6 +155,20 @@ export class CacheInfo {
     };
   }
 
+  static TokenHourChar(tokenIdentifier: string): CacheInfo {
```
bogdan-rosianu, line 158: Char -> Chart

src/utils/cache.info.ts
```diff
@@ -155,6 +155,20 @@ export class CacheInfo {
...
+    return {
+      key: `tokenHourChart:${tokenIdentifier}`,
+      ttl: Constants.oneMinute() * 10,
+    };
+  }
+
+  static TokenDailyChar(tokenIdentifier: string, after: string): CacheInfo {
```
bogdan-rosianu, line 165: Char -> Chart

src/endpoints/mex/mex.token.charts.service.ts
```diff
@@ -0,0 +1,106 @@
...
+      async () => await this.getTokenPricesDayResolutionRaw(tokenIdentifier, after),
+      CacheInfo.TokenDailyChar(tokenIdentifier, after).ttl,
+    );
+  }
+
+  async getTokenPricesDayResolutionRaw(tokenIdentifier: string, after: string): Promise<MexTokenChart[]> {
+    const tokenExists = await this.checkTokenExists(tokenIdentifier);
+
+    if (!tokenExists) {
+      this.logger.error(`Token ${tokenIdentifier} does not exist or has no trading pair`);
```
bogdan-rosianu, line 69: also here, I don't think we should log each user bad request. Also, I see that we return `[]` for this case but also for exceptions. Will the empty array (`[]`) be cached for both situations? Because for the one when a token does not exist, it's good, but when we encounter an exception while fetching data from the external service, we should not cache an empty result because of an internal issue

src/endpoints/mex/mex.token.charts.service.ts
cfaur09: fixed it, I used undefined if the given token is not part of mex tokens. Also, if undefined, from mex controller I throw NotFoundException

src/utils/cache.info.ts
cfaur09: done

src/utils/cache.info.ts
cfaur09: done

tanghel (approved):

PR #1336, opened 2024-09-19
 contracts deploys endpoints

cfaur09:
## Proposed Changes
- for the `<api>/accounts/:address/deploys` -> use `deployer` for the elastic query match
- for the `<api>/accounts/:address/contracts` -> use `currentOwner` for the elastic query match

## How to test
- `<api>/accounts/erd1ss6u80ruas2phpmr82r42xnkd6rxy40g9jl69frppl4qez9w2jpsqj8x97/deploys` -> should return only the deploys that the given address has been made
- `<api>/accounts/erd1ss6u80ruas2phpmr82r42xnkd6rxy40g9jl69frppl4qez9w2jpsqj8x97/contracts` -> should return only the contracts where the given address is owner
- `<api>/accounts/erd1ss6u80ruas2phpmr82r42xnkd6rxy40g9jl69frppl4qez9w2jpsqj8x97/deploys/count` -> should return only the deploys count that the given address has been made ( `±71` )
- `<api>/accounts/erd1ss6u80ruas2phpmr82r42xnkd6rxy40g9jl69frppl4qez9w2jpsqj8x97/contracts/count` -> should return only the contracts count where the given address is owner ( `±20`)

### DIFF
--- a/src/common/indexer/elastic/elastic.indexer.service.ts
+++ b/src/common/indexer/elastic/elastic.indexer.service.ts
@@ -51,7 +51,7 @@ export class ElasticIndexerService implements IndexerInterface {
     return await this.elasticService.getCount('operations', query);
   }
 
-  async getAccountContractsCount(address: string): Promise<number> {
+  async getAccountDeploysCount(address: string): Promise<number> {
     const elasticQuery: ElasticQuery = ElasticQuery.create()
       .withCondition(QueryConditionOptions.must, [QueryType.Match("deployer", address)]);
 
@@ -437,7 +437,7 @@ export class ElasticIndexerService implements IndexerInterface {
     );
   }
 
-  async getAccountContracts(pagination: QueryPagination, address: string): Promise<any[]> {
+  async getAccountDeploys(pagination: QueryPagination, address: string): Promise<any[]> {
     const elasticQuery: ElasticQuery = ElasticQuery.create()
       .withPagination(pagination)
       .withCondition(QueryConditionOptions.must, [QueryType.Match("deployer", address)])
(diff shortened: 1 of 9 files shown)


src/common/indexer/elastic/elastic.indexer.service.ts
```diff
@@ -446,6 +446,22 @@ export class ElasticIndexerService implements IndexerInterface {
     return await this.elasticService.getList('scdeploys', "contract", elasticQuery);
   }
 
+  async getAccountContracts(pagination: QueryPagination, address: string): Promise<any[]> {
```
bogdan-rosianu, line 449: could have used the same function and parameterize the es field to be used for query match (`fieldToMatch` that can be either `currentOwner`, either `deployer`)

src/endpoints/accounts/account.controller.ts
```diff
@@ -1091,21 +1091,47 @@ export class AccountController {
...
+  @ApiOperation({ summary: 'Account contracts details', description: 'Returns contracts details for a given account' })
   @ApiQuery({ name: 'from', description: 'Number of items to skip for the result set', required: false })
   @ApiQuery({ name: 'size', description: 'Number of items to retrieve', required: false })
   @ApiOkResponse({ type: [DeployedContract] })
   getAccountContracts(
     @Param('address', ParseAddressPipe) address: string,
     @Query('from', new DefaultValuePipe(0), ParseIntPipe) from: number,
     @Query('size', new DefaultValuePipe(25), ParseIntPipe) size: number,
-  ): Promise<DeployedContract[]> {
+  ): Promise<any[]> {
```
bogdan-rosianu, line 1129: might declare a class instead of using `any`. But the same class (`DeployedContract`) can be used. you might rename it

cfaur09: done

src/common/indexer/elastic/elastic.indexer.service.ts
cfaur09: i will leave for now the actual implementation, more refactors for elastic indexer will follow

PR #1374, opened 2024-11-07
Add 'addresses' query parameter to filter accounts by a list of addresses

cfaur09:
## Proposed Changes
- Introduces a new query parameter to filter accounts by a list of addresses and includes validation to ensure the addresses array does not exceed 25 elements.

## How to test
- `<api>/accounts?addresses=erd1an4xpn58j7ymd58m2jznr32t0vmas75egrdfa8mta6fzvqn9tkxq4jvghn,erd1kyle4p6chlvj0u6fyds7wj47jqwq6kh9h083kw8j76254l8u49dqepfz86`
- `<api>/accounts?addresses[... >25 ] -> should throw 400 Bad Request`

### DIFF
--- a/src/common/gateway/gateway.service.ts
+++ b/src/common/gateway/gateway.service.ts
@@ -94,6 +94,11 @@ export class GatewayService {
     return result;
   }
 
+  async getAccountsBulk(addresses: string[]): Promise<Account[]> {
+    const result = await this.create('address/bulk', GatewayComponentRequest.addressDetails, addresses);
+    return result.accounts;
+  }
+
   async getEsdtSupply(identifier: string): Promise<EsdtSupply> {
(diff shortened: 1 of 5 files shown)


src/common/gateway/gateway.service.ts
```diff
@@ -94,6 +94,11 @@ export class GatewayService {
     return result;
   }
 
+  async getAccountsBulk(addresses: string[]): Promise<Account[]> {
+    const result = await this.create('address/bulk', GatewayComponentRequest.addressDetails, addresses);
```
bogdan-rosianu, line 98: we should create a new component instead of using `GatewayComponentRequest.addressDetails`

cfaur09: done

bogdan-rosianu (approved):

dragos-rebegea (approved):

PR #1379, opened 2024-11-13
Add ResponseTimeCpuInterceptor to monitor response time and CPU usage per endpoint

cfaur09:
## Reasoning
- To monitor and log the response time and CPU usage for each endpoint in the application.
- To provide warnings when response time or CPU usage exceeds specified thresholds.
- To improve the observability and performance monitoring of the application.
  
## Proposed Changes
- Added a new `ResponseTimeCpuInterceptor` to log response time and CPU usage for each endpoint.
- Configured the interceptor to log warnings if response time exceeds `1000ms` or CPU usage exceeds `500ms`.
- Integrated the interceptor globally in the application.

## How to test
- Fast Endpoint:
Make a request to a fast endpoint and verify that no warnings are logged.
- Slow Endpoint:
Make a request to a slow endpoint and verify that warnings for response time and CPU usage are logged.

### DIFF
--- a/src/interceptors/response-time.cpu.interceptor.ts
+++ b/src/interceptors/response-time.cpu.interceptor.ts
@@ -1,0 +1,36 @@
+import { CallHandler, ExecutionContext, Injectable, NestInterceptor } from '@nestjs/common';
+import { Observable } from 'rxjs';
+import { tap } from 'rxjs/operators';
+import { MetricsService, PerformanceProfiler } from '@multiversx/sdk-nestjs-monitoring';
+
+@Injectable()
+export class ResponseTimeCpuInterceptor implements NestInterceptor {
+  constructor(private readonly metricsService: MetricsService) { }
+
+  intercept(context: ExecutionContext, next: CallHandler): Observable<any> {
+    const profiler = new PerformanceProfiler();
+    const startCpuUsage = process.cpuUsage();
+
+    return next.handle().pipe(
+      tap(() => {
+        const duration = profiler.stop();
+        const endCpuUsage = process.cpuUsage(startCpuUsage);
+        const cpuUsage = (endCpuUsage.user + endCpuUsage.system) / 1000;
+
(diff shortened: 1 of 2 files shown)


src/interceptors/response-time.cpu.interceptor.ts
```diff
@@ -0,0 +1,46 @@
+import { CallHandler, ExecutionContext, Injectable, NestInterceptor, Logger } from '@nestjs/common';
+import { Observable } from 'rxjs';
+import { tap } from 'rxjs/operators';
+import { MetricsService, PerformanceProfiler } from '@multiversx/sdk-nestjs-monitoring';
+
+@Injectable()
+export class ResponseTimeCpuInterceptor implements NestInterceptor {
+  private readonly logger = new Logger(ResponseTimeCpuInterceptor.name);
+  private readonly shouldLog: boolean;
```
bogdan-rosianu, line 9: we could have kept the entire `options` object as class member. but it can remain like this

src/main.ts
```diff
@@ -220,6 +221,8 @@ async function configurePublicApp(publicApp: NestExpressApplication, apiConfigSe
   globalInterceptors.push(new RequestCpuTimeInterceptor(metricsService));
   // @ts-ignore
   globalInterceptors.push(new LoggingInterceptor(metricsService));
+  //@ts-ignore
+  globalInterceptors.push(new ResponseTimeCpuInterceptor(metricsService, { shouldLog: false }));
```
bogdan-rosianu, line 225: shouldLog default on true?

cfaur09: pushed

src/interceptors/response-time.cpu.interceptor.ts
cfaur09: we can leave it like that

bogdan-rosianu (approved):

PR #1389, opened 2024-11-19
Api 58 sovereign token prefix

bogdan-rosianu:
## Reasoning
- 
- 
- 
  
## Proposed Changes
- 
- 
- 

## How to test
- 
- 
-


src/endpoints/esdt/esdt.address.service.ts
```diff
@@ -217,14 +216,14 @@ export class EsdtAddressService {
...
       const identifier = filter.identifiers[0];
 
-      if (identifier.split('-').length === 2) {
+      if (TokenUtils.isSovereignIdentifier(identifier)) {
+        collection = identifier.split('-').slice(0, 3).join('-');
+        nonceHex = identifier.split('-')[3];
+      } else {
         collection = identifier.split('-').slice(0, 2).join('-');
         nonceHex = identifier.split('-')[2];
       }
```
cfaur09, line 225: you can avoid identifier.split('-') by adding into a const like:

const parts = identifier.split('-')

src/endpoints/esdt/esdt.address.service.ts
```diff
@@ -268,7 +268,11 @@ export class EsdtAddressService {
     for (const dataSourceNft of nfts) {
       const nft = new NftAccount();
       nft.identifier = dataSourceNft.tokenIdentifier;
-      nft.collection = dataSourceNft.tokenIdentifier.split('-').slice(0, 2).join('-');
+      if (TokenUtils.isSovereignIdentifier(dataSourceNft.tokenIdentifier)) {
```
cfaur09, line 271: same as above

src/endpoints/esdt/esdt.address.service.ts
bogdan-rosianu: fixed

src/endpoints/esdt/esdt.address.service.ts
bogdan-rosianu: fixed

cfaur09 (approved):

PR #1395, opened 2024-11-25
API-98: relayed v3 transactions support

bogdan-rosianu:
## Reasoning
- new relayed v3 transaction comming from new protocol versions lack support for filtering and both fields (`relayer` and `relayerSignature`) inside the API
  
## Proposed Changes
- adjust queries/filter/responses to include the new transactions

## How to test
- `<api>/transactions/<hash of a tx with relayed v3>` should contain `relayer` and `relayerSignature`
- `<api>/transaction?relayer=erd1...` should work

### DIFF
--- a/src/common/gateway/entities/transaction.ts
+++ b/src/common/gateway/entities/transaction.ts
@@ -43,4 +43,6 @@ export class Transaction {
   logs: TransactionLog | undefined = undefined;
   receipt: TransactionReceipt | undefined = undefined;
   smartContractResults: GatewaySmartContractResults[] | undefined = undefined;
+  relayerAddress: string = '';
+  relayerSignature: string = '';
 }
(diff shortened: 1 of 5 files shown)


src/endpoints/transfers/transfer.controller.ts
```diff
@@ -162,6 +165,8 @@ export class TransferController {
     @Query('before', ParseIntPipe) before?: number,
     @Query('after', ParseIntPipe) after?: number,
     @Query('round', ParseIntPipe) round?: number,
+    @Query('relayer', ParseAddressPipe) relayer?: string,
+    @Query('isRelayed', new ParseBoolPipe) isRelayed?: boolean,
```
cfaur09, line 169: you can remove the `new` keyword if you don't add any params to ParseBoolPipe

bogdan-rosianu: fixed

dragos-rebegea (approved):

PR #1408, opened 2024-12-03
API-160: tx pool endpoint tests + new cs image

bogdan-rosianu:
## Reasoning
- e2e tests for the `pool` endpoint
  
## Proposed Changes
- unit tests + new CS image


src/test/chain-simulator/pool.cs-e2e.ts
```diff
@@ -30,20 +28,13 @@ describe('Pool e2e tests with chain simulator', () => {
         `${config.apiServiceUrl}/pool`,
       );
       const pool = response.data;
-      let numTimesAliceWasSender = 0;
-      let numTimesBobWasReceiver = 0;
       for (const tx of pool) {
-        numTimesAliceWasSender += tx.sender === config.aliceAddress ? 1 : 0;
-        numTimesBobWasReceiver += tx.receiver === config.bobAddress ? 1 : 0;
         expect(tx).toHaveProperty('receiver');
```
cfaur09, line 32: instead of using .toHaveProperty you can add all the props into array and then use for loop to check if the response contain the props

bogdan-rosianu: done

cfaur09 (approved):

PR #1425, opened 2024-12-17
added function name for pool transactions

bogdan-rosianu:
## Reasoning
- pool transactions did not have the 'function' field which would make the pool analysis easier
  
## Proposed Changes
- add function name to pool transactions if it exists

## How to test
- `/transaction/pool` should contain the transactions functions

### DIFF
--- a/src/endpoints/pool/pool.service.ts
+++ b/src/endpoints/pool/pool.service.ts
@@ -124,7 +124,8 @@ export class PoolService {
         (!filters.receiver || transaction.receiver === filters.receiver) &&
         (!filters.type || transaction.type === filters.type) &&
         (filters.senderShard === undefined || transaction.senderShard === filters.senderShard) &&
-        (filters.receiverShard === undefined || transaction.receiverShard === filters.receiverShard)
+        (filters.receiverShard === undefined || transaction.receiverShard === filters.receiverShard) &&
+        (filters.functions === undefined || transaction.function === undefined || filters.functions.indexOf(transaction.function) > -1)
       );
     });
   }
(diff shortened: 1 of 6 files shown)


src/endpoints/pool/pool.service.ts
```diff
@@ -105,6 +109,11 @@ export class PoolService {
       transaction.receiverShard = AddressUtils.computeShard(AddressUtils.bech32Decode(transaction.receiver), shardCount);
     }
 
+    const metadata = await this.transactionActionService.getTransactionMetadata(this.poolTransactionToTransaction(transaction), false);
```
andreibancioiu, line 112: In our Python implementation, we have a small issue with parsing some badly-formatted transactions. If the issue exists in the JS implementation, as well, a tiny failure might make the whole request fail, I think. Perhaps have a more pessimistic approach?

andreibancioiu: Additionally, maybe we should benchmark this? Or is the logic applied for a mere page of transactions, not all?

If applied on all currently in pool, we might have a small performance issue?

:thought_balloon:

cfaur09:
Add also the `function` filter to be able to filter by functions

src/endpoints/pool/pool.service.ts
bogdan-rosianu: > a tiny failure might make the whole request fail, I think
no, if the decoding fails, it will just returned `undefined`, therefore not applying the function name

regarding the performance impact, it is ok since we cache the entire pool so the processing only happens once

dragos-rebegea (approved):

PR #1450, opened 2025-02-13
update dockerfile env vars

bogdan-rosianu:
## Reasoning
- 
- 
- 
  
## Proposed Changes
- 
- 
- 

## How to test
- 
- 
-


bogdan-rosianu:
will close as the updated env var is already handled in #1436

PR #1481, opened 2025-04-09
fix after query param for mex/tokens/daily

cfaur09:
### Reasoning
- if `after` param was not added, `start` graphql param was undefined and graphql returns 400 Bad Request 
  
### Proposed Changes
- add `after` graphql query param only if after is defined with value

### How to test
- `<api>/mex/tokens/prices/daily/TADA-5c032c` -> should return an array of values
- `mex/tokens/prices/daily/TADA-5c032c?after=1707688800` -> should return values that begins with the specified after value 

### Reasoning
- we need to have also `before` query timestamp to be able to return multiple values before the given daily token timestamp

### Proposed Changes
- added `before` query param
- If we have `before` but no `after`, we fetch all data (since we can't filter by end date in GraphQL) and filter in memory
- If we have `after` but no `before`, we use the GraphQL start parameter to filter at the source
- If we have `both`, we fetch from `after` and filter by `before` in memory

### How to test
- `mex/tokens/prices/daily/TADA-5c032c?before=1707516000` -> should return values before the given timestamp
- `mex/tokens/prices/daily/TADA-5c032c?before=1707516000&after=1707516000` -> should return values before the given timestamp and after the give timestamp

### DIFF
--- a/src/endpoints/mex/mex.controller.ts
+++ b/src/endpoints/mex/mex.controller.ts
@@ -183,8 +183,10 @@ export class MexController {
   @ApiNotFoundResponse({ description: 'Price not available for given token identifier' })
   async getTokenPricesDayResolution(
     @Param('identifier', ParseTokenPipe) identifier: string,
-    @Query('after') after: string): Promise<MexTokenChart[] | undefined> {
-    const charts = await this.mexTokenChartsService.getTokenPricesDayResolution(identifier, after);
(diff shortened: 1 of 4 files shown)


gabrielmatei (dismissed):

cfaur09:
New PR created where we remove after query param https://github.com/multiversx/mx-api-service/pull/1483

PR #1484, opened 2025-04-11
es circuit breaker

bogdan-rosianu:
## Reasoning
- if the underlying `elasticsearch` service is under high load, we should limit the number of requests we send
  
## Proposed Changes
- added a proxy/wrapper over the elastic search handler that will apply the circuit breaker pattern over it.
- in case requests take longer than normally expected, we should block further incoming requests towards es until the response time decreases

## How to test
- 
- 
-

### DIFF
--- a/src/common/indexer/elastic/circuit-breaker/circuit.breaker.proxy.service.ts
+++ b/src/common/indexer/elastic/circuit-breaker/circuit.breaker.proxy.service.ts
@@ -1,8 +1,9 @@
 import { OriginLogger } from "@multiversx/sdk-nestjs-common";
 import { ElasticQuery, ElasticService } from "@multiversx/sdk-nestjs-elastic";
-import { ServiceUnavailableException } from "@nestjs/common";
-import { ApiConfigService } from "../../api-config/api.config.service";
+import { Injectable, ServiceUnavailableException } from "@nestjs/common";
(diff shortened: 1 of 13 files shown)


gabrielmatei (approved):

GuticaStefan (approved):

src/common/indexer/elastic/circuit-breaker/circuit.breaker.proxy.service.ts
```diff
@@ -0,0 +1,109 @@
...
+  private readonly config: { durationThresholdMs: number, failureCountThreshold: number, resetTimeoutMs: number };
+
+  constructor(
+    readonly apiConfigService: ApiConfigService,
+    private readonly elasticService: ElasticService,
+  ) {
+    this.enabled = apiConfigService.isElasticCircuitBreakerEnabled();
+    this.config = apiConfigService.getElasticCircuitBreakerConfig();
+    this.logger.log(`ES Circuit Breaker. Enabled: ${this.enabled}. Duration threshold: ${this.config.durationThresholdMs}ms. 
+    FailureCountThreshold: ${this.config.failureCountThreshold}ms. FailureCountThreshold: ${this.config.failureCountThreshold}`);
```
cfaur09, line 22: `failureCountThreshold` value is applied multiple times

PR #1492, opened 2025-05-06
Extend features flags config values

bogdan-rosianu:
## Reasoning
- not all the features activation flags were exposed via Rest API
  
## Proposed Changes
- add missing flags

## How to test
- check `localhost:3001/about` endpoint

### DIFF
--- a/src/common/api-config/api.config.service.ts
+++ b/src/common/api-config/api.config.service.ts
@@ -893,7 +893,9 @@ export class ApiConfigService {
   }
 
   getAssetsCdnUrl(): string {
-    return this.configService.get<string>('features.assetsFetch.assetesUrl') ?? 'https://tools.multiversx.com/assets-cdn';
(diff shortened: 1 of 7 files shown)


src/common/api-config/api.config.service.ts
```diff
@@ -893,7 +893,9 @@ export class ApiConfigService {
   }
 
   getAssetsCdnUrl(): string {
-    return this.configService.get<string>('features.assetsFetch.assetesUrl') ?? 'https://tools.multiversx.com/assets-cdn';
+    return this.configService.get<string>('features.assetsFetch.assetsUrl')
+      ?? this.configService.get<string>('features.assetsFetch.assetesUrl')
```
dragos-rebegea, line 897: Create a todo to remove this in the future

bogdan-rosianu: added

PR #1499, opened 2025-05-23
add support for new NFT events

cfaur09:
## Reasoning
- 
- 
- 
  
## Proposed Changes
- 
- 
- 

## How to test
- 
- 
-

### DIFF
--- a/src/common/rabbitmq/rabbitmq.nft.handler.service.ts
+++ b/src/common/rabbitmq/rabbitmq.nft.handler.service.ts
@@ -8,6 +8,7 @@ import { NotifierEvent } from './entities/notifier.event';
 import { CacheService } from "@multiversx/sdk-nestjs-cache";
 import { BinaryUtils, OriginLogger } from '@multiversx/sdk-nestjs-common';
 import { IndexerService } from '../indexer/indexer.service';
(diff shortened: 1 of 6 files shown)


dragos-rebegea (dismissed):

src/common/rabbitmq/rabbitmq.nft.handler.service.ts
```diff
@@ -105,6 +145,62 @@ export class RabbitMqNftHandlerService {
     }
   }
 
+  public async handleNftBurnEvent(event: NotifierEvent): Promise<boolean> {
+    const identifier = this.getNftIdentifier(event.topics);
+
+    this.logger.log(`Detected 'ESDTNFTBurn' event for NFT with identifier '${identifier}'`);
+
+    try {
+      await this.cachingService.delete(`nft:${identifier}`);
```
GuticaStefan, line 154: I think here we also need to use redis pub-sub service in order to emit a delete local cache key event for the rest of the api instances within the cluster

cfaur09: added

dragos-rebegea (approved):

PR #1501, opened 2025-05-28
Api 626 applications page

cfaur09:
## Reasoning
- 
- 
- 
  
## Proposed Changes
- 
- 
- 

## How to test
- 
- 
-

### DIFF
--- a/src/common/indexer/elastic/elastic.indexer.helper.ts
+++ b/src/common/indexer/elastic/elastic.indexer.helper.ts
@@ -763,6 +763,14 @@ export class ElasticIndexerHelper {
       elasticQuery = elasticQuery.withRangeFilter('timestamp', new RangeLowerThanOrEqual(filter.before));
     }
 
+    if (filter.isVerified !== undefined) {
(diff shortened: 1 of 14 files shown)


src/common/indexer/elastic/elastic.indexer.helper.ts
```diff
@@ -761,14 +761,39 @@ export class ElasticIndexerHelper {
...
+      elasticQuery = elasticQuery.withMustMultiShouldCondition(filter.addresses, address => QueryType.Match('address', address));
+    }
+
+    if (filter.search) {
+      elasticQuery = elasticQuery.withSearchWildcardCondition(filter.search, ['address']);
+    }
+
+    if (filter.isVerified !== undefined) {
+      if (filter.isVerified) {
+        elasticQuery = elasticQuery.withMustExistCondition('api_isVerified');
```
bogdan-rosianu, line 781: shouldn't we also check if it is set to `true` ? or we only have `api_isVerified: true` and nothing if `false` ?

PR #1505, opened 2025-06-03
Tokens supply format from plugin

bogdan-rosianu:
## Reasoning
- 
- 
- 
  
## Proposed Changes
- 
- 
- 

## How to test
- 
- 
-

### DIFF
--- a/src/endpoints/esdt/esdt.service.ts
+++ b/src/endpoints/esdt/esdt.service.ts
@@ -16,6 +16,7 @@ import { EsdtType } from "./entities/esdt.type";
 import { ElasticIndexerService } from "src/common/indexer/elastic/elastic.indexer.service";
 import { randomUUID } from "crypto";
 import { EsdtSubType } from "./entities/esdt.sub.type";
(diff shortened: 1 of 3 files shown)


src/endpoints/esdt/esdt.service.ts
```diff
@@ -367,7 +369,9 @@ export class EsdtService {
   }
 
   async getTokenSupply(identifier: string): Promise<EsdtSupply> {
-    const { supply, minted, burned, initialMinted } = await this.gatewayService.getEsdtSupply(identifier);
+    const esdtSupply = await this.gatewayService.getEsdtSupply(identifier);
+    this.pluginService.formatTokenSupply(identifier, esdtSupply);
```
dragos-rebegea, line 373: this should not fail if the formatTokenSupply is not defined in the plugin module

bogdan-rosianu: yes, just tested and it works

dragos-rebegea (approved):

PR #1506, opened 2025-06-03
500ms cronjob pace

bogdan-rosianu:
## Reasoning
- 
- 
- 
  
## Proposed Changes
- 
- 
- 

## How to test
- 
- 
-


bogdan-rosianu:
I will close this PR because a better option would be to use `@Interval(500)` decorator

Issue #1151, opened 2023-11-22
[Bug]: Potential minor bug on `/accounts/{address}/delegation` endpoint


silvestrst-crypto:
### Description

Potential minor bug on `/accounts/{address}/delegation` endpoint

### Steps to reproduce

![image](https://github.com/multiversx/mx-api-service/assets/103181555/097904c2-421e-40ba-94c7-4422407312da)

Field `userUndelegatedList` is marked as required, however, when the list is empty it seems that deserialization for that field is skipped:
https://api.elrond.com/accounts/<insert an account without any active undelegations>/delegation

It might be that this issue was not hit before, but recently the response deserialization failed (rust serde) due to absence of the required field.

It's a very minor point, as the fix is simple on my side, but decided to flag it anyway.

### On which network is the bug manifesting itself?

Mainnet

### Relevant log output

```shell
error decoding response body: missing field `userUndelegatedList` at line 1 column 253
```
```


### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cfaur09:
Hi, thanks for reporting this, we will fix this issue :pray:

PR #1507, opened 2025-06-04
fix userUndelegatedList field

cfaur09:
## Reasoning
- https://github.com/multiversx/mx-api-service/issues/1151
  
## Proposed Changes
- mark userUndelegatedList as [] if is missing

## How to test
- <api>/accounts/:address/delegation -> use an address that have some undelegated EGLD

### DIFF
--- a/src/endpoints/delegation/delegation.service.ts
+++ b/src/endpoints/delegation/delegation.service.ts
@@ -74,7 +74,7 @@ export class DelegationService {
 
       return data.map((delegation: any) => new AccountDelegation({
         ...delegation,
-        userUndelegatedList: delegation.userUndelegatedList || [],
+        userUndelegatedList: delegation.userUndelegatedList ?? [],
       }));
     } catch (error) {
       this.logger.error(`Error when getting account delegation details for address ${address}`);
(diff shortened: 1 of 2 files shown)


src/endpoints/delegation/delegation.service.ts
```diff
@@ -71,7 +71,11 @@ export class DelegationService {
   async getDelegationForAddress(address: string): Promise<AccountDelegation[]> {
     try {
       const { data } = await this.apiService.get(`${this.apiConfigService.getDelegationUrl()}/accounts/${address}/delegations`);
-      return data;
+
+      return data.map((delegation: any) => new AccountDelegation({
+        ...delegation,
+        userUndelegatedList: delegation.userUndelegatedList || [],
```
GuticaStefan, line 77: maybe here would be more useful to use '??' instead of '||' if we want to return '[]' only in the cases when `userUndelegatedList` is null | undefined ?

cfaur09: good point

bogdan-rosianu (approved):

dragos-rebegea (approved):

PR #1528, opened 2025-08-20
Websockets 

stefangutica:


src/endpoints/blocks/blocks.gateway.ts
```diff
@@ -0,0 +1,48 @@
...
+
+@WebSocketGateway({ cors: { origin: '*' } })
+export class BlocksGateway implements OnGatewayDisconnect {
+    @WebSocketServer()
+    server!: Server;
+
+    constructor(private readonly blockService: BlockService) { }
+
+    @SubscribeMessage('subscribeBlocks')
+    async handleSubscription(client: Socket, payload: any) {
```
cfaur09: no validation on incoming websocket payloads and any malformed data can crash the service

src/endpoints/blocks/blocks.gateway.ts
```diff
@@ -0,0 +1,48 @@
...
+        shard: filter.shard,
+        proposer: filter.proposer,
+        validator: filter.validator,
+        epoch: filter.epoch,
+        nonce: filter.nonce,
+        hashes: filter.hashes,
+        order: filter.order,
+      });
+
+      const blocks = await this.blockService.getBlocks(
```
cfaur09, line 37: add error handling (try/catch) for the getBlocks to be sure that we handle the case of an error

src/endpoints/network/network.gateway.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  constructor(private readonly networkService: NetworkService) { }
+
+  @SubscribeMessage('subscribeStats')
+  async handleSubscription(client: Socket) {
+    await client.join('statsRoom');
+  }
+
+  async pushStats() {
+    if (this.server.sockets.adapter.rooms.has('statsRoom')) {
```
cfaur09, line 18: maybe we need to improve the logic here because even if the room is empty, always fetches the stats so we need to check if the room contains more than 1, then fetch the date from getStats

src/endpoints/blocks/blocks.gateway.ts
stefangutica: added

src/endpoints/network/network.gateway.ts
stefangutica: the rooms are handled by socket.io library, the convention is that a room exists only if it has at least 1 client inside

src/crons/websocket/blocks.gateway.ts
```diff
@@ -0,0 +1,70 @@
...
+  server!: Server;
+
+  constructor(private readonly blockService: BlockService) { }
+
+  @SubscribeMessage('subscribeBlocks')
+  async handleSubscription(
+    @ConnectedSocket() client: Socket,
+    @MessageBody(new WsValidationPipe()) payload: BlockSubscribePayload
+  ) {
+    const filterHash = JSON.stringify(payload);
```
bogdan-rosianu: this is not really a hash, so maybe we should update the naming to something like `filterIdentifier`. Please apply this suggestion in all the gateways

src/endpoints/pool/entities/pool.subscribe.ts
```diff
@@ -0,0 +1,20 @@
+// block-subscribe.dto.ts
```
bogdan-rosianu: remove this comment

src/utils/ws-validation.pipe.ts
```diff
@@ -0,0 +1,16 @@
+// ws-validation.pipe.ts
```
bogdan-rosianu: remove this comment

stefangutica: fixed

src/endpoints/blocks/entities/block.subscribe.ts
```diff
@@ -0,0 +1,29 @@
+// block-subscribe.dto.ts
```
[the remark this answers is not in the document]
stefangutica: fixed

src/crons/websocket/connection.handler.ts
```diff
@@ -0,0 +1,20 @@
...
+
+    @WebSocketServer()
+    server!: Server;
+
+    afterInit(__server: Server) { }
+
+    handleDisconnect(_client: Socket) { }
+
+    handleConnection(client: Socket, ..._args: any[]) {
+        client.setMaxListeners(12);
```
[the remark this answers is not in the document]
stefangutica: if the number is exceeded we will only receive a warning(not error) from EventEmitter, but it doesn't affect the functionality. For the moment, until we add another class like blocks,pools,... for ws subscription, that warning should not appear

PR #1532, opened 2025-09-10
State changes

stefangutica:


.github/workflows/chain-simulator-e2e-tests.yml
```diff
@@ -47,6 +50,7 @@ jobs:
 
       - run: npm ci
       - run: npm run init
+      - run: cat src/common/indexer/entities/account.ts
```
bogdan-rosianu: this can be removed (only used for debugging)

src/common/api-config/api.config.service.ts
```diff
@@ -972,4 +976,49 @@ export class ApiConfigService {
...
+    const stateChangesPort = this.configService.get<number>('features.stateChanges.port');
+    if (stateChangesPort === undefined) {
+      throw new Error('No state changes port present');
+    }
+
+    return stateChangesPort;
+  }
+
+  getStateChangesUrl(): string {
+    const url = this.configService.get<string>('features.stateChanges.url');
```
bogdan-rosianu: maybe we should provide a better naming, such as `features.stateChanges.rabbitUrl` ? add a fallback to `features.stateChanges.url` as well

src/endpoints/accounts-v2/entities/account.ts
```diff
@@ -0,0 +1,54 @@
+import { SwaggerUtils } from "@multiversx/sdk-nestjs-common";
+import { ApiProperty } from "@nestjs/swagger";
+import { AccountAssets } from "src/common/assets/entities/account.assets";
+
+export class Account {
```
bogdan-rosianu: all the entities seem to be duplicates of the v1 entities. or are there differences?

src/state-changes/state.changes.consumer.service.ts
```diff
@@ -0,0 +1,284 @@
...
+
+    constructor(
+        private readonly cacheService: CacheService,
+        private readonly accountDetailsRepository: AccountDetailsRepository,
+        @Inject('PUBSUB_SERVICE') private clientProxy: ClientProxy,
+    ) { }
+
+    @CompetingRabbitConsumer({
+        exchange: 'state_accesses',
+        queueName: 'api_state_accesses_queue-test',
```
bogdan-rosianu: is there a way to provide these from config? not sure if the decorator supports variables

config/config.devnet.yaml
```diff
@@ -26,8 +26,10 @@ features:
   stateChanges:
     enabled: false
     port: 5675
-    url: 'amqp://guest:guest@127.0.0.1:5672'
+    rabbitUrl: 'amqp://guest:guest@127.0.0.1:5672'
```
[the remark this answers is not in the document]
stefangutica: done

PR #1534, opened 2025-09-15
improve getBlocks

cfaur09:
## Reasoning
* Sequential async operations in `getBlocks` method causing performance bottlenek
* Sequential processing of blocks via computeProposerAndValidators calls
* Sequential identity lookups in applyProposerIdentity method

## Proposed Changes
- Replace sequential block processing with Promise.all() for parallel execution
- Optimize applyProposerIdentity to fetch all identities in parallel
- Pre-filter relevant nodes before async operations to reduce unnecessary calls

## How to test
- Call `/blocks` endpoint with various pagination sizes (e.g., 25 50 blocks)

### DIFF
--- a/src/endpoints/blocks/block.service.ts
+++ b/src/endpoints/blocks/block.service.ts
@@ -34,8 +34,8 @@ export class BlockService {
 
   async getBlocks(filter: BlockFilter, queryPagination: QueryPagination, withProposerIdentity?: boolean): Promise<Block[]> {
     const result = await this.indexerService.getBlocks(filter, queryPagination);
-    const blocks = [];
-    for (const item of result) {
+
+    const blocks = await Promise.all(result.map(async (item) => {
       const blockRaw = await this.computeProposerAndValidators(item);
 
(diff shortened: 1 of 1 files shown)


src/endpoints/blocks/block.service.ts
```diff
@@ -58,17 +58,16 @@ export class BlockService {
...
-      const nodeIdentity = node.identity;
-      if (!nodeIdentity) {
-        continue;
-      }
+    const identityPromises = relevantNodes.map(async (node) => {
+      const identity = await this.identitiesService.getIdentity(node.identity!);
+      return { node, identity };
+    });
+
+    const nodeIdentities = await Promise.all(identityPromises);
```
bogdan-rosianu, line 68: we currently have 171 identities on mainnet. I don't think ES or other underlying services will handle this concurrent load that well. we can use a sort of concurrency limiting solution, such as [p-limit](https://doziestar.medium.com/serving-tasks-efficiently-understanding-p-limit-in-javascript-fb524a35b846), Bluebird or something custom.

stefangutica (approved):

PR #1570, opened 2025-12-03
Add support for timestampMs for filtering

stefangutica:
## Reasoning
- starting with supernova, we must provide support for filtering by timestampMs
  
## Proposed Changes
- accept timestampMs as paramter for `before` and `after` filtering + add timestampMs as optional response field where is the case

## How to test
- you can filter by `timestamp` or `timestampMs` in `before` and `after` fields and the response should be correct


src/utils/time.utils.ts
```diff
@@ -0,0 +1,5 @@
+export class TimeUtils {
+  static isTimestampInSeconds(input: number): boolean {
+    return input < 100_000_000_000;
```
cfaur09: no handling of negative numbers or zero. add input <= 0 
just to be sure

cfaur09: add 100_000_000_000 into static readonly const

src/common/indexer/elastic/elastic.indexer.service.ts
```diff
@@ -1031,8 +1032,9 @@ export class ElasticIndexerService implements IndexerInterface {
   }
 
   async getBlockByTimestampAndShardId(timestamp: number, shardId: number): Promise<Block | undefined> {
+    const timestampIdentifier = TimeUtils.isTimestampInSeconds(timestamp) ? 'timestamp' : 'timestampMs';
     const elasticQuery = ElasticQuery.create()
-      .withRangeFilter('timestamp', new RangeGreaterThanOrEqual(timestamp))
+      .withRangeFilter(timestampIdentifier, new RangeGreaterThanOrEqual(timestamp))
       .withCondition(QueryConditionOptions.must, [QueryType.Match('shardId', shardId, QueryOperator.AND)])
       .withSort([{ name: 'timestamp', order: ElasticSortOrder.ascending }]);
```
cfaur09, line 1039: .withSort()
can we sort after timestampMs?

src/utils/time.utils.ts
stefangutica: done

stefangutica: added in timestamp parser pipe

config/config.testnet.yaml
```diff
@@ -200,3 +200,7 @@ compression:
   level: 6
   threshold: 1024
   chunkSize: 16384
+supernova:
```
bogdan-rosianu: please refactor the config:
```
chainSupernova:
  enabled: false
  activationEpoch: 2030
  activationTimestamp: 1864749472
```

and place it right under `stakingV5`

src/common/api-config/api.config.service.ts
```diff
@@ -1067,4 +1067,22 @@ export class ApiConfigService {
 
     return undefined;
   }
+
+  getSupernovaActivationEpoch(): number {
+    const epoch = this.configService.get<number>('supernova.activation.epoch');
+    if (epoch == null) {
+      throw new Error('No supernova.activation.epoch present');
```
bogdan-rosianu: please let's not make API crash if new configs are missing. refactor the config as recommended and check if the enabled flag exists and return false otherwise

src/common/indexer/elastic/elastic.indexer.service.ts
bogdan-rosianu: I don't think so because not all the records have this field, so we'll need to use `timestamp` which should be available for all of them and it's fine

bogdan-rosianu (dismissed):

cfaur09 (dismissed):

bogdan-rosianu (approved):

cfaur09 (approved):

PR #1583, opened 2026-01-22
Supernova execution results mb

bogdan-rosianu:
## Reasoning
- starting with supernova, the miniblock details + other items should be fetched from a new Elasticsearch index named `executionresults`
  
## Proposed Changes
- fetch new data from the new index
- merge the 2 objects (the execution results block has a better prio)
- add support for gateway enable epochs endpoints
- check if supernova flag exists and is activated

### DIFF
--- a/src/endpoints/blocks/block.service.ts
+++ b/src/endpoints/blocks/block.service.ts
@@ -131,7 +131,7 @@ export class BlockService {
         result.proposer = publicKeys[result.proposer];
       }
       if (!isChainAndromedaEnabled) {
-        result.validators = result.validators.map((validator: number) => publicKeys[validator]);
+        result.validators = result.validators?.map((validator: number) => publicKeys[validator]);
       } else {
         result.validators = publicKeys;
       }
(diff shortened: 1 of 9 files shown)


stefangutica (approved):

src/endpoints/blocks/block.service.ts
```diff
@@ -109,11 +112,20 @@ export class BlockService {
...
 
   async getBlock(hash: string): Promise<BlockDetailed> {
-    const result = await this.indexerService.getBlock(hash) as any;
+    let result = await this.indexerService.getBlock(hash) as any;
 
     const isChainAndromedaEnabled = this.apiConfigService.isChainAndromedaEnabled()
       && result.epoch >= this.apiConfigService.getChainAndromedaActivationEpoch();
 
+    const supernovaEnableEpoch = await this.getSupernovaEnableEpoch();
+    const isSupernovaEnabled = supernovaEnableEpoch !== -1 && result.epoch >= supernovaEnableEpoch;
```
cfaur09, line 121: to avoid complexity, I recomand to simplify the code like:

```
if (supernovaEnableEpoch === -1 || result.epoch < supernovaEnableEpoch) {
  return result;
}

const executionResults = await this.indexerService.getExecutionResults(hash);
if (!executionResults) {
  return result;
}

return ApiUtils.mergeObjects(result, executionResults);
```

bogdan-rosianu: I cannot return since there are more instructions left in the function that will use the `result` variable

cfaur09 (approved):

PR #1611, opened 2026-06-30
Filter Token Accounts by Account Type

stefangutica:
## Reasoning
- token holders accounts endpoint returned all holders,  normal wallets holders + smart contracts holders
 
## Proposed Changes
- add filtering option depending on account type

## How to test
- when filtering by an account type, only that specific account type should be returned as a response

### DIFF
--- a/src/common/indexer/elastic/elastic.indexer.service.ts
+++ b/src/common/indexer/elastic/elastic.indexer.service.ts
@@ -17,7 +17,7 @@ import { TokenFilter } from "src/endpoints/tokens/entities/token.filter";
 import { Block } from "../entities/block";
 import { Tag } from "../entities/tag";
 import { ElasticIndexerHelper } from "./elastic.indexer.helper";
-import { TokenType } from "../entities";
+import { AccountType, TokenType } from "../entities";
 import { SortCollections } from "src/endpoints/collections/entities/sort.collections";
 import { AccountQueryOptions } from "src/endpoints/accounts/entities/account.query.options";
 import { AccountSort } from "src/endpoints/accounts/entities/account.sort";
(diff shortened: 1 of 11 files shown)


src/common/indexer/elastic/elastic.indexer.service.ts
```diff
@@ -139,21 +139,36 @@ export class ElasticIndexerService implements IndexerInterface {
...
   }
 
-  async getTokenAccountsCount(identifier: string): Promise<number | undefined> {
-    const elasticQuery: ElasticQuery = ElasticQuery.create()
+  async getTokenAccountsCount(identifier: string, accountType?: AccountType): Promise<number | undefined> {
+    let elasticQuery: ElasticQuery = ElasticQuery.create()
       .withCondition(QueryConditionOptions.must, [QueryType.Match("token", identifier, QueryOperator.AND)]);
 
+
+    const contractAddressPrefix = 'erd1' + 'q'.repeat(12);;
```
claudiulataretu: Extract this logic into a helper function. Same code used in the `getTokenAccounts` method

claudiulataretu (changes requested):

src/common/indexer/elastic/elastic.indexer.service.ts
claudiulataretu: stray double semicolon and double blank line. const contractAddressPrefix = 'erd1' + 'q'.repeat(12);; has a trailing ;;, and lines 145-146 are two consecutive blank lines

claudiulataretu (changes requested):

src/common/indexer/elastic/elastic.indexer.service.ts
stefangutica: added

claudiulataretu (approved):

popenta (approved):

